PR #101, opened 2021-11-07
[PAPI-240] Token history endpoints

drmathias:
* New token history endpoints
* Custom `DateTime` model binder that always binds to a UTC date - based on https://github.com/dotnet/aspnetcore/issues/11584
* `Interval` enum for snapshot interval selection, uses `EnumMember` attributes so the API can accept values beginning with a number (`1H`, `1D` ...) . Required a type converter and couple more attributes to work with model binding.
* `DateTime` pointer is encoded as a unix epoch time, otherwise encodes with `:` which screws up cursor parsing. Be aware this _could_ be a problem elsewhere.


src/Opdex.Platform.Application.Abstractions/EntryQueries/MarketTokens/GetMarketTokenSnapshotsWithFilterQuery.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// Creates a request to retrieve snapshot data from a market for a given token.
+        /// </summary>
+        /// <param name="token">The address of the token.</param>
+        /// <param name="market">The address of the market.</param>
+        /// <param name="cursor">The snapshot cursor filter.</param>
+        public GetMarketTokenSnapshotsWithFilterQuery(Address market, Address token, SnapshotCursor cursor)
+        {
+            if (market == Address.Empty) throw new ArgumentNullException(nameof(market));
+            if (token == Address.Empty) throw new ArgumentNullException(nameof(token));
+            if (cursor is null) throw new ArgumentNullException(nameof(cursor));
```
mrtpain, line 24: Nit - as we create queries/handlers we should add tests for these validations w/ exception messages. 

Personally, I often small minor fixes necessary for these when writing tests whether the exception thrown is the wrong type, or the message I had was incorrect.

src/Opdex.Platform.Application.Abstractions/Models/Tokens/TokenSnapshotDto.cs
```diff
@@ -5,11 +5,7 @@ namespace Opdex.Platform.Application.Abstractions.Models.Tokens
 {
     public class TokenSnapshotDto
```
mrtpain, line 6: I think we should add the token address for clients to reference if necessary, what do you think?

src/Opdex.Platform.Application.Abstractions/Queries/Tokens/Snapshots/RetrieveTokenSnapshotsWithFilterQuery.cs
```diff
@@ -1,64 +1,36 @@
...
-            if (endDate.Equals(default) || endDate < startDate)
-            {
-                throw new ArgumentOutOfRangeException(nameof(endDate));
-            }
-
-            if (!snapshotType.IsValid())
-            {
-                throw new ArgumentOutOfRangeException(nameof(snapshotType));
-            }
+            if (marketId < 0) throw new ArgumentOutOfRangeException(nameof(marketId), "Market id must be greater or equal to zero.");
```
mrtpain, line 24: This can probably be removed now that `marketId` is `ulong` I think the original check for `< 0` was when we were using `long` for Ids.

src/Opdex.Platform.Infrastructure.Abstractions/Data/Queries/Tokens/Snapshots/SelectTokenSnapshotsWithFilterQuery.cs
```diff
@@ -1,64 +1,35 @@
...
-            if (endDate.Equals(default) || endDate < startDate)
-            {
-                throw new ArgumentOutOfRangeException(nameof(endDate));
-            }
-
-            if (!snapshotType.IsValid())
-            {
-                throw new ArgumentOutOfRangeException(nameof(snapshotType));
-            }
+            if (marketId < 0) throw new ArgumentOutOfRangeException(nameof(marketId), "Market id must be greater or equal to zero.");
```
mrtpain, line 23: Same comment here around this check

src/Opdex.Platform.WebApi/Models/Requests/SnapshotFilterParameters.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        /// <summary>
+        /// Start time for which to retrieve snapshots.
+        /// </summary>
+        public DateTime StartTime { get; set; }
+
+        /// <summary>
+        /// End time for which to retrieve snapshots.
+        /// </summary>
+        public DateTime EndTime { get; set; }
```
mrtpain, line 22: Probably nothing, curious on if you think `StartDateTime` and `EndDateTime` might be clearer for clients, probably doesn't matter at all, just a thought.

test/Opdex.Platform.Application.Tests/EntryHandlers/MarketTokens/GetMarketTokenSnapshotsWithFilterQueryHandlerTests.cs
```diff
@@ -0,0 +1,406 @@
...
+using Opdex.Platform.Infrastructure.Abstractions.Data.Queries;
+using System;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Opdex.Platform.Application.Tests.EntryHandlers.MarketTokens
+{
+    public class GetMarketTokenSnapshotsWithFilterQueryHandlerTests
```
mrtpain, line 28: ahhh these are the fun tests 💀

test/Opdex.Platform.WebApi.Tests/Validation/SnapshotFilterParametersValidatorTests.cs
```diff
@@ -0,0 +1,160 @@
+using FluentValidation.TestHelper;
+using Opdex.Platform.Infrastructure.Abstractions.Data.Queries;
+using Opdex.Platform.WebApi.Models.Requests;
+using Opdex.Platform.WebApi.Validation;
+using System;
+using Xunit;
+
+namespace Opdex.Platform.WebApi.Tests.Validation
+{
+    public class SnapshotFilterParametersValidatorTests
```
mrtpain, line 10: Another thought for future tasks we should probably come back to these and assert the messages we want are returned.

src/Opdex.Platform.WebApi/Controllers/TokensController.cs
```diff
@@ -97,33 +98,23 @@ public async Task<ActionResult<TokenResponseModel>> GetToken([FromRoute] Address
...
-        /// <returns><see cref="TokenSnapshotHistoryResponseModel"/> with a list of historical data points.</returns>
+        /// <param name="filters">Filter parameters.</param>
+        /// <param name="cancellationToken">Cancellation token.</param>
+        /// <returns>Paged token snapshot data.</returns>
         [HttpGet("{address}/history")]
-        [ProducesResponseType(typeof(TokenSnapshotHistoryResponseModel), StatusCodes.Status200OK)]
+        [ProducesResponseType(typeof(TokenSnapshotsResponseModel), StatusCodes.Status200OK)]
+        [ProducesResponseType(typeof(ValidationProblemDetails), StatusCodes.Status400BadRequest)]
         [ProducesResponseType(typeof(ProblemDetails), StatusCodes.Status401Unauthorized)]
         [ProducesResponseType(typeof(ProblemDetails), StatusCodes.Status404NotFound)]
```
mrtpain, line 110: For future though what do you think about not returning 404's on filter related requests? Alternatively, returning empty result set instead. So when market || token are not found, just exit early instead. 

Wouldn't be for this task, I think all of our filter endpoints are operating this way with 404's but might be worth ignoring and returning empty results.

mrtpain (commented):
LGTM for the most part, couple comments for consideration awesome work as always 🥳

src/Opdex.Platform.WebApi/Controllers/TokensController.cs
drmathias: Disagree with this, the 404 status code is the right way to handle it. Guessing you're suggesting this in relation to exception logs, in which case I think we shouldn't use, or should suppress exceptions so they're not logged.

src/Opdex.Platform.Application.Abstractions/Queries/Tokens/Snapshots/RetrieveTokenSnapshotsWithFilterQuery.cs
drmathias: Cool, removed these checks

src/Opdex.Platform.WebApi/Models/Requests/SnapshotFilterParameters.cs
drmathias: I agree, makes it clearer. Updated them

src/Opdex.Platform.Application.Abstractions/Models/Tokens/TokenSnapshotDto.cs
drmathias: Not in the individual token snapshot but in the page result, sure. It doesn't seem that useful to me honestly, as if you're making a request you have to already have the token and market address. Although I see no harm in adding it. I'll amend the `TokenSnapshotsResponseModel` and create a `MarketTokenSnapshotsResponseModel` so we can add the market address in the market tokens response as well.

test/Opdex.Platform.WebApi.Tests/Validation/SnapshotFilterParametersValidatorTests.cs
drmathias: Added reason validation and created a backlog item for the existing validation tests

src/Opdex.Platform.Application.Abstractions/EntryQueries/MarketTokens/GetMarketTokenSnapshotsWithFilterQuery.cs
drmathias: IMO, testing isn't that important here and the exception message is unnecessary where we're null checking. The property name provides enough info and can be tested.  but I agree if we can we should test

drmathias: Added some tests and messages where we're checking Address == Empty

mrtpain: Tests validating at least the property including in the exception is good enough for me, its more just about making sure the _right_ exception is thrown according to expectations.

mrtpain (approved):

PR #102, opened 2021-11-08
[PAPI-170] IP rate limiting

drmathias:
Basic IP rate limiting. Allows us to configure rules for each API instance.
Config can be used to turn off rate limiting for specific IP ranges, which could be useful, for example, if you wanted to route traffic through an API gateway.


mrtpain (approved):
Nice!

PR #103, opened 2021-11-09
[PAPI-88] Liquidity Pool History

mrtpain:
Implements pagination on the get liquidity pool history endpoint following the get token history implementation.

*adding more test coverage but otherwise ready for review*

### DIFF
--- a/src/Opdex.Platform.Infrastructure/Data/Handlers/Tokens/Snapshots/SelectTokenSnapshotsWithFilterQueryHandler.cs
+++ b/src/Opdex.Platform.Infrastructure/Data/Handlers/Tokens/Snapshots/SelectTokenSnapshotsWithFilterQueryHandler.cs
@@ -34,8 +34,8 @@ namespace Opdex.Platform.Infrastructure.Data.Handlers.Tokens.Snapshots
             WHERE
                 {nameof(TokenSnapshotEntity.TokenId)} = @{nameof(SqlParams.TokenId)} AND
                 {nameof(TokenSnapshotEntity.MarketId)} = @{nameof(SqlParams.MarketId)} AND
-                {nameof(TokenSnapshotEntity.StartDate)} BETWEEN @{nameof(SqlParams.StartDate)} AND @{nameof(TokenSnapshotEntity.EndDate)} AND
-                {nameof(TokenSnapshotEntity.SnapshotTypeId)} = @{nameof(TokenSnapshotEntity.SnapshotTypeId)}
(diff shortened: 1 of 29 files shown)


src/Opdex.Platform.WebApi/Models/Requests/SnapshotFilterParameters.cs
```diff
@@ -25,7 +25,7 @@ public class SnapshotFilterParameters : FilterParameters<SnapshotCursor>
         protected override SnapshotCursor InternalBuildCursor()
         {
             if (Cursor is null) return new SnapshotCursor(Interval, StartDateTime, EndDateTime, Direction, Limit, PagingDirection.Forward, default);
-            Base64Extensions.TryBase64Decode(Cursor, out var decodedCursor);
+            Cursor.TryBase64Decode(out var decodedCursor);
```
drmathias, line 28: nitpick: Don't feel too strongly about it in this particular case, but my preference is to avoid extension methods on common types like `string`.

src/Opdex.Platform.Infrastructure/Data/Handlers/Tokens/Snapshots/SelectTokenSnapshotsWithFilterQueryHandler.cs
```diff
@@ -116,7 +117,7 @@ private static string QueryBuilder(SelectTokenSnapshotsWithFilterQuery request)
         }
 
         // consider better solution, likely need to reuse
-        private SnapshotType ConvertToSnapshotType(Interval interval)
+        private static SnapshotType ConvertToSnapshotType(Interval interval)
```
drmathias, line 120: nitpic: Something I avoided initially as wasn't sure if it would be reused. Could be worth creating a static method in a separate class so we're not duplicating this everywhere.

drmathias (commented):
Nice, looks good to merge when that ASC/DESC conflict is sorted out

src/Opdex.Platform.WebApi/Models/Requests/SnapshotFilterParameters.cs
mrtpain: Easier to read though imo, if the extension is there idk that it hurts to use it the best way we can simply because of the underlying type but 🤷‍♂️ nbd I can revert

src/Opdex.Platform.Infrastructure/Data/Handlers/Tokens/Snapshots/SelectTokenSnapshotsWithFilterQueryHandler.cs
mrtpain: I see how it is, put it on the next guy huh haha

PR #105, opened 2021-11-12
Attempt Indexer Unlock

mrtpain:
Attempt at resolving the occasional indexer locking seen on Testnet.

If the indexer is locked and the indexer comes back around to process next blocks, check if the instance that locked the table is the instance attempting to index, if so, unlock and move forward.

Based on findings of this implementation, we can take it further with a rewind to ensure no data was missed.

### DIFF
--- a/src/Opdex.Platform.WebApi/IndexerBackgroundService.cs
+++ b/src/Opdex.Platform.WebApi/IndexerBackgroundService.cs
@@ -59,8 +59,18 @@ namespace Opdex.Platform.WebApi
 
                     if (indexLock.Locked)
                     {
-                        _logger.LogWarning(IndexingAlreadyRunningLog);
-                        continue;
+                        if (indexLock.InstanceId != _opdexConfiguration.InstanceId)
+                        {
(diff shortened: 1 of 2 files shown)


src/Opdex.Platform.WebApi/IndexerBackgroundService.cs
```diff
@@ -59,8 +59,18 @@ protected override async Task ExecuteAsync(CancellationToken cancellationToken)
...
                     if (indexLock.Locked)
                     {
-                        _logger.LogWarning(IndexingAlreadyRunningLog);
-                        continue;
+                        if (indexLock.InstanceId != _opdexConfiguration.InstanceId)
+                        {
+                            _logger.LogWarning(IndexingAlreadyRunningLog);
+                            continue;
+                        }
+                        else
```
drmathias, line 67: nitpick: Else shouldn't be needed because of the `continue;`

drmathias: Resolved in my PR onto this branch

PR #107, opened 2021-11-13
[PAPI-247] Deploy Fixes

drmathias:
* Correctly (de)serialize byte array smart contract params
* Remove auth from deployment - you can only have a market address of the auth token once the contracts are deployed

### DIFF
--- a/src/Opdex.Platform.WebApi/Controllers/DeployController.cs
+++ b/src/Opdex.Platform.WebApi/Controllers/DeployController.cs
@@ -1,5 +1,4 @@
 using MediatR;
-using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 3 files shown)


src/Opdex.Platform.WebApi/Controllers/DeployController.cs
```diff
@@ -52,7 +51,6 @@ public DeployController(IMediator mediator, ILogger<DeployController> logger, Op
         /// <param name="cancellationToken">cancellation token.</param>
         /// <returns>Success</returns>
         [HttpPost("dev-contracts")]
-        [Authorize(Policy = "AdminOnly")]
```
mrtpain, line 55: This needs to come back

drmathias: Cool, didn't realise the market address wasn't required to auth with an admin key. Re-added it.

PR #108, opened 2021-11-15
[PAPI-244] Wallet position improvements

drmathias:
Replaces `IncludeLpTokens` filter with `TokenType` that allows balances to be filtered by: `NonProvisional`, `Provisional` and `All`.

PR #109, opened 2021-11-16
[PAPI-231] Query params schema

drmathias:
* Adds minimum/maximum to integer query params
* Cleans up nullable usage
* Mark required query params
* Renamed Provisional query param to TokenType to be consistent

Fairly ugly using attributes, although unfortunately it's not an easy feat to do it any other way. Also limited to what we can define. For example, we cannot really define regex patterns for addresses, as attributes are setup with constant values and the address format would have to be resolved based on the network config.

Only concern is NSwag generates minimum/maximum with a decimal point, so `1.0` as an example. _Shouldn't_ be a problem as it's defined integer but something to watch out for.

### DIFF
--- a/src/Opdex.Platform.WebApi/Models/Requests/FilterParameters.cs
+++ b/src/Opdex.Platform.WebApi/Models/Requests/FilterParameters.cs
@@ -1,6 +1,8 @@
+using Microsoft.AspNetCore.Mvc;
+using NJsonSchema.Annotations;
 using Opdex.Platform.Common.Enums;
 using Opdex.Platform.Infrastructure.Abstractions.Data.Queries;
-using System;
+using System.ComponentModel.DataAnnotations;
 
 namespace Opdex.Platform.WebApi.Models.Requests
 {
@@ -19,12 +21,16 @@ namespace Opdex.Platform.WebApi.Models.Requests
         /// <summary>
         /// Number of results to return per page.
         /// </summary>
-        public uint Limit { get; set; }
+        // virtual is used to be able to override the attribute
+        [Range(1, Cursor.DefaultMaxLimit)]
+        public virtual uint Limit { get; set; }
 
         /// <summary>
         /// The cursor when paging.
         /// </summary>
-        public string Cursor { get; set; }
+        [NotNull]
+        [FromQuery(Name = "cursor")]
+        public string EncodedCursor { get; set; }
 
         /// <summary>
         /// Builds a cursor filter used for querying.
--- a/src/Opdex.Platform.WebApi/Models/Requests/FilterParameters.cs
+++ b/src/Opdex.Platform.WebApi/Models/Requests/FilterParameters.cs
@@ -29,7 +29,7 @@ namespace Opdex.Platform.WebApi.Models.Requests
         /// The cursor when paging.
         /// </summary>
         [NotNull]
-        [FromQuery(Name = "cursor")]
+        [FromQuery(Name = "Cursor")]
         public string EncodedCursor { get; set; }
 
         /// <summary>
--- a/src/Opdex.Platform.WebApi/Models/Requests/Tokens/TokenFilterParameters.cs
+++ b/src/Opdex.Platform.WebApi/Models/Requests/Tokens/TokenFilterParameters.cs
@@ -1,3 +1,4 @@
+using NJsonSchema.Annotations;
 using Opdex.Platform.Common.Extensions;
 using Opdex.Platform.Common.Models;
 using Opdex.Platform.Infrastructure.Abstractions.Data.Queries;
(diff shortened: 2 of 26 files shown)


src/Opdex.Platform.WebApi/Models/Requests/FilterParameters.cs
```diff
@@ -19,12 +21,16 @@ public abstract class FilterParameters<TCursor> where TCursor : Cursor
...
+        [Range(1, Cursor.DefaultMaxLimit)]
+        public virtual uint Limit { get; set; }
 
         /// <summary>
         /// The cursor when paging.
         /// </summary>
-        public string Cursor { get; set; }
+        [NotNull]
+        [FromQuery(Name = "Cursor")]
+        public string EncodedCursor { get; set; }
```
mrtpain, line 33: We could probably use a base class here for these filter requests to mitigate the repetitiveness and going forward take care of the shared properties in one place.

src/Opdex.Platform.WebApi/Models/Requests/Tokens/TokenFilterParameters.cs
```diff
@@ -21,23 +22,25 @@ public TokenFilterParameters()
...
         /// Tokens to filter specifically for.
         /// </summary>
+        [NotNull]
         public IEnumerable<Address> Tokens { get; set; }
 
         /// <summary>
         /// A generic keyword search against token addresses, names and ticker symbols.
         /// </summary>
+        [NotNull]
         public string Keyword { get; set; }
```
mrtpain, line 37: These two can be nullable

src/Opdex.Platform.WebApi/Models/Requests/FilterParameters.cs
mrtpain: Also the cursor can be not null

src/Opdex.Platform.WebApi/Models/Requests/Transactions/TransactionFilterParameters.cs
```diff
@@ -23,17 +24,19 @@ public TransactionFilterParameters()
...
         /// Optional list of smart contract addresses to filter transactions by.
         /// </summary>
+        [NotNull]
         public IEnumerable<Address> Contracts { get; set; }
 
         /// <summary>
         /// Filter transactions based on event types included.
         /// </summary>
+        [NotNull]
         public IEnumerable<TransactionEventType> EventTypes { get; set; }
```
mrtpain, line 34: Can be nullable

src/Opdex.Platform.WebApi/Models/Requests/Wallets/AddressBalanceFilterParameters.cs
```diff
@@ -18,6 +19,7 @@ public AddressBalanceFilterParameters()
         /// <summary>
         /// Specific tokens to lookup.
         /// </summary>
+        [NotNull]
         public IEnumerable<Address> Tokens { get; set; }
```
mrtpain, line 23: Nullable

src/Opdex.Platform.WebApi/Models/Requests/Wallets/MiningPositionFilterParameters.cs
```diff
@@ -19,11 +20,13 @@ public MiningPositionFilterParameters()
...
         /// The specific mining pools to include.
         /// </summary>
+        [NotNull]
         public IEnumerable<Address> MiningPools { get; set; }
 
         /// <summary>
         /// The specific liquidity pools to include.
         /// </summary>
+        [NotNull]
         public IEnumerable<Address> LiquidityPools { get; set; }
```
mrtpain, line 30: Nullable

src/Opdex.Platform.WebApi/Models/Requests/Wallets/StakingPositionFilterParameters.cs
```diff
@@ -17,6 +17,7 @@ public StakingPositionFilterParameters()
         /// <summary>
         /// The specific liquidity pools to include.
         /// </summary>
+        [NotNull]
```
mrtpain, line 20: Nullable

mrtpain:
Couple comments around Nullability, otherwise LGTM.

Don't merge just yet though until there is a matching PR to line up the new `TokenFilter` rename.

src/Opdex.Platform.WebApi/Models/Requests/Transactions/TransactionFilterParameters.cs
drmathias: OpenAPI 3.0 has definitions for both nullable and required. Nullable would mean you can pass null explicitly, for example in the json body. For reference types NSwag generates the doc with nullable true. Here `[NotNull]` is removing the nullable definition from the OpenAPI doc, because the API doesn't accept a null query string.

mrtpain: I'll have to look into that one, that sounds incredibly wrong and misleading.

drmathias: Take a look at the top answer here https://stackoverflow.com/questions/45575493/what-does-required-in-openapi-really-mean

To give an example, `?cursor=null` would be treated as the string "null" by asp.net core. NSwag adds `nullable: true` to everything which is a reference type, it doesn't determine whether it is a query param or not.

src/Opdex.Platform.WebApi/Models/Requests/FilterParameters.cs
drmathias: Not sure I understand? This class is a base class for those properties

src/Opdex.Platform.WebApi/Models/Requests/Transactions/TransactionFilterParameters.cs
mrtpain: Yeah I see required != not null - confusing at first but makes sense

src/Opdex.Platform.WebApi/Models/Requests/FilterParameters.cs
mrtpain: 👀 ....... 💀

mrtpain:
PR to merge and deploy along with this, for the frontend: https://github.com/Opdex/opdex-ui/pull/59

drmathias:
@mrtpain Will leave you to merge and deploy this, unsure if #108 could require a front-end change as well

PR #110, opened 2021-11-16
[PAPI-149] Minimum and maximum

drmathias:
Specifies minimum and maximum on number fields for OpenAPI. Use of `double.MaxValue` indicates NSwag to ignore the maximum. Pretty ugly but it's the only way.

### DIFF
--- a/src/Opdex.Platform.WebApi/Models/Requests/Index/RewindRequest.cs
+++ b/src/Opdex.Platform.WebApi/Models/Requests/Index/RewindRequest.cs
@@ -11,6 +11,7 @@ namespace Opdex.Platform.WebApi.Models.Requests.Index
         /// The block number to rewind too.
         /// </summary>
         [Required]
+        [Range(1, double.MaxValue)]
         public ulong Block { get; set; }
     }
 }
--- a/src/Opdex.Platform.WebApi/Models/Requests/Markets/CreateStandardMarketQuoteRequest.cs
+++ b/src/Opdex.Platform.WebApi/Models/Requests/Markets/CreateStandardMarketQuoteRequest.cs
@@ -12,7 +12,7 @@ namespace Opdex.Platform.WebApi.Models.Requests.Markets
         public Address MarketOwner { get; set; }
 
         [Required]
-        [Range(0, double.MaxValue)]
+        [Range(0, 10)]
         public uint TransactionFee { get; set; }
 
         [Required]
--- a/src/Opdex.Platform.WebApi/Models/Requests/Markets/CreateStandardMarketQuoteRequest.cs
+++ b/src/Opdex.Platform.WebApi/Models/Requests/Markets/CreateStandardMarketQuoteRequest.cs
@@ -12,6 +12,7 @@ namespace Opdex.Platform.WebApi.Models.Requests.Markets
         public Address MarketOwner { get; set; }
 
         [Required]
+        [Range(0, double.MaxValue)]
         public uint TransactionFee { get; set; }
 
         [Required]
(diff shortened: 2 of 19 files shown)


src/Opdex.Platform.WebApi/Models/Requests/Markets/CreateStandardMarketQuoteRequest.cs
```diff
@@ -12,6 +12,7 @@ public class CreateStandardMarketQuoteRequest
         public Address MarketOwner { get; set; }
 
         [Required]
+        [Range(0, double.MaxValue)]
         public uint TransactionFee { get; set; }
```
mrtpain, line 16: 10 is max here

mrtpain:
I question in general whether we should display these on response models because we won't for FixedDecimal types right?

src/Opdex.Platform.WebApi/Models/Requests/Index/RewindRequest.cs
```diff
@@ -11,6 +11,7 @@ public class RewindRequest
         /// The block number to rewind too.
         /// </summary>
         [Required]
+        [Range(1, double.MaxValue)]
         public ulong Block { get; set; }
```
mrtpain, line 15: should this be ulong.MaxValue for block related values?

drmathias: NSwag interprets `double.MaxValue` as not defining a maximum. `ulong.MaxValue` defines ulong maxValue as maximum. Confusing I know, sadly there's no separate min/max attributes. We just want to leave it out because in the OpenApi doc it already gets defined as int64.

drmathias:
> I question in general whether we should display these on response models because we won't for FixedDecimal types right?

For completeness, though I doubt it is important to have them. FixedDecimal is defined string in OpenApi so minimum/maximum is not valid.

src/Opdex.Platform.WebApi/Models/Requests/Index/RewindRequest.cs
mrtpain: Wouldn't  we want to have `(1, ulong.MaxValue)` rather than 1 to infinite for these since they're block related and we know the range?

drmathias: Your comment led me down the rabbit hole. Essentially, we're seeing `ulong` values generate schema as
```
"block": {
  "type": "integer",
  "format": "int64",
  "minimum": 1.0
}
```

However, this is wrong as the OpenAPI spec describes the format int64 as:

> Signed 64-bit integers (long type).

So I've opened RicoSuter/NJsonSchema/pull/1437 to hopefully change this.

And so I think there's two possibilities to improve this, either:

1. We add a maximum like you suggest
2. Adding a custom format

Existing OpenAPI tools will not understand a custom format, we'd have to modify them to be able to interpret it. Adding a maximum _may_ be understood by existing tools. We could try to do both.


Awkwardly, everyone seems to handle long/ulong differently. [Google](https://developers.google.com/discovery/v1/type-format) provides `int64` and `uint64` as a string and defines custom formats for `uint32` and `uint64` values. Twitter sends both a number and a string in its API. Apparently JS [doesn't support 64 bit integers fully](https://stackoverflow.com/a/34989371). Could this become a problem for our front end?

Maybe it's more a problem for legacy code, if so my take is that we could use a number and include a custom format. Not sure how difficult a custom format would be with NSwag however. Then begs the question, maybe we should be doing the same for uint256, serialising it as a number with a custom format.

src/Opdex.Platform.WebApi/Models/Requests/Markets/CreateStandardMarketQuoteRequest.cs
drmathias: Good spot, updated

src/Opdex.Platform.WebApi/Models/Requests/Index/RewindRequest.cs
mrtpain: Personally, I don't think we should put effort into it for such a small benefit since our ranges are 0 to Max for those types. 

Let's just pin it for right now, I don't think the benefits of these implementations on 0-Max values is worth it.

mrtpain (approved):

PR #114, opened 2021-11-18
[PAPI-245] Stratis Open Auth Protocol

drmathias:
- [x] Define callback
- [x] Signed message verification
- [x] UID verification
- [x] Creation of bearer token
- [x] WebSocket token push


src/Opdex.Platform.WebApi/Controllers/AuthController.cs
```diff
@@ -24,6 +33,50 @@ public AuthController(AuthConfiguration authConfiguration, IMediator mediator)
...
+        /// <response code="200">Signature was validated successfully.</response>
+        /// <response code="400">The request is not valid.</response>
+        [HttpPost]
+        [ProducesResponseType(StatusCodes.Status200OK)]
+        [ProducesResponseType(StatusCodes.Status400BadRequest)]
+        public async Task<IActionResult> StratisOpenAuthCallback([FromQuery] StratisOpenAuthCallbackQuery query,
+                                                                 [FromBody] StratisOpenAuthCallbackBody body, CancellationToken cancellationToken)
+        {
+            var expectedCallbackPath = System.IO.Path.Combine(_authConfiguration.StratisOpenAuthProtocol.CallbackBase, AuthPath);
+            var expectedId = new StratisId(expectedCallbackPath, query.Uid, query.Exp);
```
mrtpain, line 52: OpdexConfiguration has the API URL callback base per network, but will need the https:// stripped out. Would be better to reuse that vs a new config.

That config also has `WalletTransactionCallback` set to `/transactions`.

Consider doing something similar like that here so the auth config points to `/auth` vs the const string in the controller should this logic ever move outside of the controller.

drmathias: Updated to reuse config and retrieve callback path from config

mrtpain (approved):
LGTM

Can you update our actual docs on README with the change from this readme and also add a link to that mac fix 👍

PR #123, opened 2021-11-26
[PAPI-237] .NET 6

drmathias:
* Bare minimum migration to .NET 6
* Removed custom UTC model binder, see https://github.com/dotnet/aspnetcore/pull/24893/files

From the .NET docs:
> You may need to delete the bin and obj folders. Run dotnet nuget locals --clear all to clear the NuGet package cache.


src/Opdex.Platform.Common/Extensions/ListExtensions.cs
```diff
@@ -1,23 +0,0 @@
...
-    {
-        /// <summary>
-        /// Split an IEnumerable into multiple smaller chunks.
-        /// </summary>
-        /// <param name="items">The items to split up.</param>
-        /// <param name="itemsPerChunk">The number of items per chunked list.</param>
-        /// <typeparam name="T">The type of IEnumerable being chunked.</typeparam>
-        /// <returns>2 dimensional array of chunked lists.</returns>
-        public static IEnumerable<IEnumerable<T>> Chunk<T>(this IEnumerable<T> items, int itemsPerChunk)
-        {
```
mrtpain, line 16: We use this is a few areas, curious how removing this works?

drmathias: .NET 6 LINQ has this method natively https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.chunk?view=net-6.0

mrtpain:
Thinking now, will this need bicep changes for the app service before we can deploy?

drmathias:
Don't believe so

PR #126, opened 2021-11-30
[PAPI-266] Index lock reason

drmathias:
* Adds Reason column to index lock
* Checks Reason is Index before attempting to force unlock if same instance has come around

### DIFF
--- a/src/Opdex.Platform.Application/EntryHandlers/Blocks/CreateRewindToBlockCommandHandler.cs
+++ b/src/Opdex.Platform.Application/EntryHandlers/Blocks/CreateRewindToBlockCommandHandler.cs
@@ -46,8 +46,7 @@ namespace Opdex.Platform.Application.EntryHandlers.Blocks
                 throw new ArgumentOutOfRangeException(nameof(request.Block), "Rewind request exceeds maximum rewind limit.");
             }
 
-            using var cancellationTokenSource = new CancellationTokenSource(TimeSpan.FromMinutes(1));
-            var rewound = await _mediator.Send(new MakeRewindToBlockCommand(request.Block), cancellationTokenSource.Token);
+            var rewound = await _mediator.Send(new MakeRewindToBlockCommand(request.Block), CancellationToken.None);
             if (!rewound) return false;
 
             _logger.LogTrace("Beginning to refresh stale records.");
--- a/src/Opdex.Platform.Application/EntryHandlers/Blocks/CreateRewindToBlockCommandHandler.cs
+++ b/src/Opdex.Platform.Application/EntryHandlers/Blocks/CreateRewindToBlockCommandHandler.cs
@@ -46,23 +46,24 @@ namespace Opdex.Platform.Application.EntryHandlers.Blocks
                 throw new ArgumentOutOfRangeException(nameof(request.Block), "Rewind request exceeds maximum rewind limit.");
             }
 
-            var rewound = await _mediator.Send(new MakeRewindToBlockCommand(request.Block));
+            using var cancellationTokenSource = new CancellationTokenSource(TimeSpan.FromMinutes(1));
+            var rewound = await _mediator.Send(new MakeRewindToBlockCommand(request.Block), cancellationTokenSource.Token);
             if (!rewound) return false;
 
             _logger.LogTrace("Beginning to refresh stale records.");
 
             // refresh stale records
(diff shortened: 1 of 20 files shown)


src/Opdex.Platform.Application/EntryHandlers/Blocks/CreateRewindToBlockCommandHandler.cs
```diff
@@ -46,23 +46,24 @@ public async Task<bool> Handle(CreateRewindToBlockCommand request, CancellationT
                 throw new ArgumentOutOfRangeException(nameof(request.Block), "Rewind request exceeds maximum rewind limit.");
             }
 
-            var rewound = await _mediator.Send(new MakeRewindToBlockCommand(request.Block));
+            using var cancellationTokenSource = new CancellationTokenSource(TimeSpan.FromMinutes(1));
```
mrtpain, line 49: I don't know if this really benefits us setting a minute long token. It might cause more problems then it fixes since a cancellation here would cause bad data issues.

drmathias: If it gets cancelled I believe it'll rollback, just like it would if it were to fail. The idea is to prevent it hanging. Happy to remove it though, as the indexer should come around and force unlock itself with an alert, so it'll be evident something is wrong.

src/Opdex.Platform.Infrastructure/Data/Handlers/Indexer/PersistIndexerLockCommandHandler.cs
```diff
@@ -1,38 +1,61 @@
...
-        private static readonly string SqlQuery =
-            @$"UPDATE index_lock
-               SET
+    private static readonly string SqlQuery =
+        @$"UPDATE index_lock
+                SET
                     {nameof(IndexLockEntity.Locked)} = 1,
                     {nameof(IndexLockEntity.ModifiedDate)} = UTC_TIMESTAMP(),
-                    {nameof(IndexLockEntity.InstanceId)} = @{nameof(IndexLockEntity.InstanceId)}
+                    {nameof(IndexLockEntity.InstanceId)} = @{nameof(SqlParams.InstanceId)}
```
mrtpain, line 21: missing comma at the end of this line

src/Opdex.Platform.Application/EntryHandlers/Blocks/CreateRewindToBlockCommandHandler.cs
mrtpain: I think we should pull this out for now.

- Wed have to adjust MakeRewindToBlockComand/SelectRewindToBlockCommand to pass the cancellation token all the way down
- We'd have to do the same for all of the other commands below this comment and all children they call and so on, thats a big chain.
- Then we could still run into a situation if a valid rewind does take more than a minute, stored procedure deletes back to block, then have half of the below commands actually finish processing, leaving us in a bad state requiring another rewind.

src/Opdex.Platform.WebApi/IndexerBackgroundService.cs
```diff
@@ -10,127 +10,147 @@
...
-                            unavailable = true;
-                            continue;
-                        }
+                        var stuckDuringRewind = indexLock.Reason == IndexLockReason.Rewind && indexLock.ModifiedDate.AddMinutes(1) < DateTime.UtcNow;
+                        var isSameInstanceReprocessing = indexLock.Reason == IndexLockReason.Index && indexLock.InstanceId == _opdexConfiguration.InstanceId;
 
-                        if (indexLock.Locked)
+                        var totalSecondsLocked = DateTime.UtcNow.Subtract(indexLock.ModifiedDate).TotalSeconds;
+
+                        if (!stuckDuringRewind && !isSameInstanceReprocessing)
```
mrtpain, line 74: What do you think about just logging errors w/ alerts for stuck rewinds over a minute? 

I'm a bit nervous about forcing an unlock around this area and it might be good to see how often it would happen prior to us just letting it automatically go wild.

drmathias: Good idea, I'll make sure we've got an alert set up for this

mrtpain (approved):

PR #127, opened 2021-12-01
[PAPI-269] Vault Domain and Entity Models

mrtpain:

PR #128, opened 2021-12-01
[PAPI-272] New vault log models

drmathias:


drmathias:
Build is failing because it depends on enum types in #127

PR #132, opened 2021-12-05
[PAPI-276] Persistance log amendments

drmathias:
* Ensure index unlock failure logged once, always. No need to throw or return as there is no way to handle and recover. Much easier to monitor now as can monitor the one log.
* Resolves Serilog middleware logging handled exceptions as 500 errors mistakenly (middleware reordering). Exception mapping now occurs before Serilog request logging
* Index lock now doesn't always throw, instead returns. No need for exception control flow; the exception is useful only to be thrown for HTTP response mapping middleware, otherwise can be handled by the caller, like in the indexing background service.
* Query failure exceptions now propagate and get handled upward. Indexer or middleware will handle logging these.
* Persist command exceptions are logged in the handlers with useful custom properties
* Re-added exception logs from response mapping middleware. Serilog middleware does not pick up exception, watching https://github.com/serilog/serilog-aspnetcore/issues/270 for solution. Have to deal with duplicate exception logs here for now.


src/Opdex.Platform.Infrastructure/Data/Handlers/Blocks/ExecuteRewindToBlockCommandHandler.cs
```diff
@@ -32,9 +32,9 @@ public async Task<bool> Handle(ExecuteRewindToBlockCommand request, Cancellation
 
             return true;
         }
-        catch (Exception)
+        catch (Exception ex)
         {
-            _logger.LogError("Failure rewinding to block {Block}.", request.Block);
+            _logger.LogError(ex, "Failure rewinding to block {Block}.", request.Block);
```
mrtpain, line 37: Could use a missing "$" here to be `$"Failure rewinding to block {Block}."`

src/Opdex.Platform.Application/Handlers/Indexer/MakeIndexerUnlockCommandHandler.cs
```diff
@@ -8,26 +7,17 @@
 
 namespace Opdex.Platform.Application.Handlers.Indexer;
 
-public class MakeIndexerUnlockCommandHandler : IRequestHandler<MakeIndexerUnlockCommand, Unit>
+public class MakeIndexerUnlockCommandHandler : AsyncRequestHandler<MakeIndexerUnlockCommand>
```
mrtpain, line 10: Doesn't really matter but you could stick with IRequestHandler here and return `Unit.Value` instead 🤷‍♂️

src/Opdex.Platform.Infrastructure/Data/Handlers/Indexer/PersistIndexerUnlockCommandHandler.cs
```diff
@@ -21,18 +22,27 @@ public class PersistIndexerUnlockCommandHandler : IRequestHandler<PersistIndexer
...
+        try
+        {
+            var command = DatabaseQuery.Create(SqlQuery, new { InstanceId = _instanceId }, CancellationToken.None);
+            var result = await _context.ExecuteCommandAsync(command);
+            if (result == 0) throw new NoRowsAffectedException("Indexer appears to have not unlocked.");
+        }
+        catch (Exception ex)
+        {
+            _logger.LogCritical(ex, "Failed to unlock indexer.");
+        }
```
mrtpain, line 46: You may have a reason for throwing and immediately catching, but it looks to me that you could just log here instead of throwing, catching, then logging.

drmathias: The intent was to follow the same log message and be able to see the cause of failure in the form of the exception. Thinking about it again, I agree the exception is not necessary.

PR #135, opened 2021-12-08
[PAPI-280] Rewind New Vault

mrtpain:
Rewind capabilities for new vaults including their contracts, proposals, votes, pledges and certificates.

Could use more tests and comments on queries/commands prior to merge.


src/Opdex.Platform.Infrastructure/Data/Handlers/Transactions/SelectTransactionForVaultProposalVoteRewindQueryHandler.cs
```diff
@@ -0,0 +1,101 @@
...
+                tl.{nameof(TransactionLogEntity.Contract)},
+                tl.{nameof(TransactionLogEntity.Details)}
+            FROM transaction t
+            LEFT JOIN transaction_log tl
+                ON t.{nameof(TransactionEntity.Id)} = tl.{nameof(TransactionLogEntity.TransactionId)}
+            WHERE t.`{nameof(TransactionEntity.From)}` = @{nameof(SqlParams.Voter)} AND
+                  tl.{nameof(TransactionLogEntity.Contract)} = @{nameof(SqlParams.Vault)} AND
+                  tl.{nameof(TransactionLogEntity.LogTypeId)} IN @{nameof(SqlParams.LogTypes)} AND
+                  JSON_EXTRACT(tl.{nameof(TransactionLogEntity.Details)}, '$.proposalId') = @{nameof(SqlParams.ProposalId)}
+            ORDER BY t.{nameof(TransactionEntity.Block)} DESC;".RemoveExcessWhitespace();
```
drmathias, line 48: Is it possible to instead have this as two separate queries? One to retrieve the transaction then one to retrieve the log. Might improve performance.

src/Opdex.Platform.Infrastructure/Data/Handlers/Transactions/SelectTransactionForVaultProposalPledgeRewindQueryHandler.cs
```diff
@@ -0,0 +1,101 @@
...
+
+public class SelectTransactionForVaultProposalPledgeRewindQueryHandler
+    : IRequestHandler<SelectTransactionForVaultProposalPledgeRewindQuery, Transaction>
+{
+    private const string Split = "Split";
+
+    private static readonly string SqlQuery =
+        @$"SELECT
+                t.{nameof(TransactionEntity.Id)},
+                t.`{nameof(TransactionEntity.To)}`,
```
drmathias, line 27: nitpick: Occasional ` character in all of these queries

drmathias (commented):
LGTM. A little concerned about performance if we were to have multiple proposals and especially vaults. Would be useful if we had a way to benchmark rewinds.

Seems to be a couple conflicts coming from my last merge, so apologies for that.

drmathias (approved):

src/Opdex.Platform.Infrastructure/Data/Handlers/Transactions/SelectTransactionForVaultProposalPledgeRewindQueryHandler.cs
mrtpain: the ` has to stay there for reserved SQL words.

src/Opdex.Platform.Infrastructure/Data/Handlers/Transactions/SelectTransactionForVaultProposalVoteRewindQueryHandler.cs
mrtpain: It'll end up being the same query because we have to filter by log and proposalId anyways, might as well use Dapper to build in one, due to the join Limit 1 won't be exact since it would pull one log for a transaction technically. I can set a lower limit like 5 or so to help out. 

I didn't enjoy making this query but its needed because the contract only stores the active balance of a pledge/vote. It doesn't keep the official vote/pledge total for a user in State, only logs.

PR #138, opened 2021-12-09
[PAPI-287] Inject network to app settings

drmathias:


drmathias:
Merging immediately to test pipeline

PR #139, opened 2021-12-09
[PAPI-290] Indexer configuration toggle

drmathias:


src/Opdex.Platform.WebApi/IndexerBackgroundService.cs
```diff
@@ -141,16 +148,16 @@ public override async Task StopAsync(CancellationToken cancellationToken)
...
-    public override Task StartAsync(CancellationToken cancellationToken)
-    {
-        return base.StartAsync(cancellationToken);
-    }
 }
+
+public class IndexerConfiguration
+{
+    public bool Enabled { get; set; }
+}
```
mrtpain, line 163: We should continue to make separate file/classes for configurations so things don't get lost within other fies.

src/Opdex.Platform.WebApi/IndexerBackgroundService.cs
```diff
@@ -120,18 +114,31 @@ protected override async Task ExecuteAsync(CancellationToken cancellationToken)
                 }
                 finally
                 {
-                    started = true;
+                    // delay 30 seconds when indexing services are unavailable
+                    // delay 8 to 12 seconds when indexing is available
+                    var seconds = unavailable ? 30 : new Random().Next(8, 13);
+                    await Task.Delay(TimeSpan.FromSeconds(seconds), cancellationToken);
```
mrtpain, line 120: This can affect shutdowns but idk if its a huge deal since the important operations should have finished.

We allow 30 second timeout on shutdown, really that may be the problem when taking polly + retries into consideration, maybe it should be more like several minutes to make sure shutdown isn't forced in the middle of anything.

I don't think this code particularly is problematic but the comment is relevant

mrtpain: Maybe a simple fix for this specifically actually
```
if (!cancellationToken.IsCancellationRequested) { ... } 
```

src/Opdex.Platform.WebApi/IndexerBackgroundService.cs
```diff
@@ -45,21 +48,12 @@ protected override async Task ExecuteAsync(CancellationToken cancellationToken)
...
-                    var seconds = unavailable ? 30 : new Random().Next(8, 13);
-
-                    await Task.Delay(TimeSpan.FromSeconds(seconds), cancellationToken);
-                }
-
                 try
                 {
                     var indexLock = await mediator.Send(new RetrieveIndexerLockQuery(), cancellationToken);
-                    if (!indexLock.Available)
+                    if (!indexLock.Available || !_indexerOptions.CurrentValue.Enabled)
```
mrtpain, line 54: Do we need `CurrentValue`?

src/Opdex.Platform.WebApi/IndexerBackgroundService.cs
drmathias: In this case we need to use `IOptionsMonitor` and so I think it's best to keep separate, so we're not monitoring config which doesn't need to be monitored. I'll move into a separate file

src/Opdex.Platform.WebApi/IndexerBackgroundService.cs
drmathias: Yeah, this is because we're using `IOptionsMonitor` which updates with config changes. Without it, we'd load config once and it would never change, even if we updated the values in the config source

src/Opdex.Platform.WebApi/IndexerBackgroundService.cs
drmathias: Not sure I understand. As we're passing the cancellation token into `Task.Delay` I'm not sure there is a difference, as the cancellation will throw `TaskCancelledException` and the delay is over. The indexer *should* be locked at this time anyway.

As a side note, I thought about shutdown and the cancellation token. Inside `StopAsync` I've now moved the call to `base.StopAsync` up to the top, as this is what requests cancellation of the token. My thinking is we are better off cancelling any potential indexing before force unlocking the indexer.

mrtpain: We want to do everything we can to shut down gracefully. The snippet I sent will not hit Task.Delay if we wrap it in that if block preventing unwanted exceptions from being thrown.

I think its better if we move the stop back into the the finally block of that try catch, basically to make sure we've finished the attempt prior to actually calling base.Stop. Imo its better to order that way even for a slight advantage rather than the stopping first potentially running into issues related with the attempted check and unlock.

drmathias: I've reverted the `Task.Delay` so it's now called inside the try block with the started flag enabled. Also added a catch block for `TaskCancelledException` so we can log something more specific. 

As for the `StopAsync()` call, I don't understand what issues could potentially occur. I believe calling `base.StopAsync()` after attempting to unlock opens up a slim possibility of the indexer being unlocked then locked once again, as the cancellation token is not cancelled until `base.StopAsync()` is called. However, this is covered by the try-finally block that the call to process latest blocks is wrapped in, so I'm happy to revert it back as well.

mrtpain: I suppose `base.StopAsync()` could be called first. Either way, theres a slight chance of edge cases for us if we're doing anything related to locking/unlocking.

I think the changes you made look good and if you feel that putting `StopAsync` first is the better move, do it 👍

mrtpain (approved):
Check the latest comment to our discussion I wrote, otherwise lgtm. Nice work as usual

PR #141, opened 2021-12-10
Socket Fix w/ URI Creation

mrtpain:
- Attempt to resolve URI creation within PlatformHub.
- Remove unused old Signalr Dependency
- WalletProvider tweak to handle nulls w/out NullRefException

### DIFF
--- a/src/Opdex.Platform.Infrastructure/Clients/SignalR/PlatformHub.cs
+++ b/src/Opdex.Platform.Infrastructure/Clients/SignalR/PlatformHub.cs
@@ -4,7 +4,6 @@ using Opdex.Platform.Common.Encryption;
 using Opdex.Platform.Common.Extensions;
 using SSAS.NET;
 using System;
-using System.IO;
 
 namespace Opdex.Platform.Infrastructure.Clients.SignalR;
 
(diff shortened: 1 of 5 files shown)


src/Opdex.Platform.Infrastructure/Clients/SignalR/PlatformHub.cs
```diff
@@ -16,8 +15,9 @@ public class PlatformHub : Hub<IPlatformClient>
     public PlatformHub(ITwoWayEncryptionProvider twoWayEncryptionProvider, AuthConfiguration authConfiguration, OpdexConfiguration opdexConfiguration)
     {
         _twoWayEncryptionProvider = twoWayEncryptionProvider ?? throw new ArgumentNullException(nameof(twoWayEncryptionProvider));
-        var authCallbackUri = new Uri(Path.Combine(opdexConfiguration.ApiUrl, authConfiguration.StratisSignatureAuth.CallbackPath));
-        _authCallback = $"{authCallbackUri.Authority}{authCallbackUri.AbsolutePath}";
+        var created = Uri.TryCreate($"{opdexConfiguration.ApiUrl}/{authConfiguration.StratisSignatureAuth.CallbackPath}", UriKind.Absolute, out var uri);
+        if (!created) throw new Exception("Unable to create callback URI");
+        _authCallback = $"{uri.Authority}{uri.AbsolutePath}";
```
drmathias, line 20: Believe that `Uri.TryCreate` will only work as expected if `opdexConfiguration.ApiUrl` doesn't contain a path. As it will trim any path on the first parameter. But I'm not sure this is going to be a problem.

Can you do the same for the endpoint in the auth controller? That builds the callback URL with `Path.Combine` as well

PR #143, opened 2021-12-14
[PAPI-299] Return vault governance TokensLocked

drmathias:
Adds a `TokensLocked` property to the vault governance response


mrtpain (approved):
👍

PR #146, opened 2021-12-15
[PAPI-301] DocumentProcessor to transform allOf to ref

drmathias:
Enables ReadMe to parse and display paginated response models
See https://docs.opdex.com/reference/transactions_gettransactions for results


mrtpain (approved):
Cool that was a fast turnaround :)

PR #147, opened 2021-12-15
[PAPI-268] Lookup transaction for notification broadcast

drmathias:


src/Opdex.Platform.Application/Handlers/Transactions/RetrieveCirrusUnverifiedTransactionSenderByHashQueryHandler.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        _mediator = mediator ?? throw new ArgumentNullException(nameof(mediator));
+    }
+
+    public async Task<Address> Handle(RetrieveCirrusUnverifiedTransactionSenderByHashQuery request, CancellationToken cancellationToken)
+    {
+        var transaction = await _mediator.Send(new CallCirrusGetRawTransactionQuery(request.TransactionHash), cancellationToken);
+
+        if (transaction is null || transaction.Vout.Length != 1
+            && transaction.Vout[0].ScriptPubKey.Addresses.Length != 1) return Address.Empty;
```
mrtpain, line 25: Do you mean `||` here instead of `&&`?

src/Opdex.Platform.Application/Handlers/Transactions/RetrieveCirrusExistsInMempoolQueryHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+        int attempt = 0;
+
+        // Retry up to MaxRetries times with the configured Backoff in seconds in between attempts
+        while (++attempt <= MaxRetries && !cancellationToken.IsCancellationRequested)
+        {
+            if (attempt > 1) await Task.Delay(TimeSpan.FromSeconds(Backoff), cancellationToken);
+
+            bool existsInMempool = await _mediator.Send(new CallCirrusGetExistsInMempoolQuery(request.TransactionHash), cancellationToken);
+
+            if (existsInMempool) return true;
```
mrtpain, line 32: we may want to try catch this so we retry total failures too, what do you think?

src/Opdex.Platform.Application/Handlers/Transactions/RetrieveCirrusUnverifiedTransactionSenderByHashQueryHandler.cs
drmathias: Good point, amended and added tests

src/Opdex.Platform.Application/Handlers/Transactions/RetrieveCirrusExistsInMempoolQueryHandler.cs
drmathias: Not sure there's any benefit to it, since the endpoint will always return 200 unless there's critical failure i.e. the node is down. Transient failures should already get handled by polly policies

mrtpain: You're right

mrtpain (approved):

PR #150, opened 2021-12-16
[PAPI-302] Transaction controller route examples and response documentation

drmathias:


drmathias:
I've removed `[Authorize]` on `GET transactions` and `GET transactions/{hash}`. Not sure if I can remove without side effects for the UI so let me know and I can add them back.

mrtpain (approved):

PR #152, opened 2021-12-19
[PAPI-297] GET /vault-governances/{address}/certificates endpoint

drmathias:


src/Opdex.Platform.Infrastructure.Abstractions/Data/Queries/VaultGovernances/Certificates/VaultGovernanceCertificatesCursor.cs
```diff
@@ -0,0 +1,94 @@
...
+        if (!Base64Extensions.TryBase64Decode(encoded, out var decoded) || !ulong.TryParse(decoded, out var result)) return false;
+
+        pointer = result;
+        return true;
+    }
+}
+
+public enum VaultCertificateStatusFilter
+{
+    All, Redeemed, Revoked
```
mrtpain, line 93: We might want to display all non-redeemed certificates.

Maybe we add a `Vesting` status here and in the select
```
WHERE certificates.Redeemed = false
```

mrtpain (approved):
Just the one comment, I think more often than not we'll only show non redeemed certificates so adding a `Vesting` filter should resolve that, otherwise lgtm

src/Opdex.Platform.Infrastructure.Abstractions/Data/Queries/VaultGovernances/Certificates/VaultGovernanceCertificatesCursor.cs
drmathias: Nice idea, added

PR #154, opened 2021-12-20
[PAPI-171] Further load testing and performance tests

drmathias:
* Stress test for `GET /liquidity-pools`, used for evaluating performance degradation
* Added first CD performance test, runs after deployment to testnet

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -72,12 +72,12 @@ stages:
           tags: '$(TAG)'
         displayName: Docker
 
-      - task: CopyFiles@2
-        inputs:
-          SourceFolder: '$(Build.SourcesDirectory)/test/k6/ci'
-          Contents: '**'
-          TargetFolder: '$(Build.ArtifactStagingDirectory)/k6'
-        displayName: Copy Load Tests Files
+#      - task: CopyFiles@2
+#        inputs:
+#          SourceFolder: '$(Build.SourcesDirectory)/test/k6/ci'
+#          Contents: '**'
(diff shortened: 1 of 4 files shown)


test/k6/get-liquidity-pools.js
```diff
@@ -0,0 +1,31 @@
+/*
+* Stress test for GET /liquidity-pools with filters.
+* Created due to anomalous 10-20s response times seen on other tests.
```
mrtpain, line 3: 😢

azure-pipelines.yml
```diff
@@ -94,7 +94,16 @@ stages:
...
+  - job:
+    displayName: 'Load Testing'
+    steps:
+      - task: k6-load-test@0
+        inputs:
+          filename: 'test/k6/ci/status/get-status.js'
+          args: '-e api_base_url=https://app-opdex-test-api-staging.azurewebsites.net'
+        displayName: 'GET /status'
+    dependsOn: 'test' # template deployment name == parameters.environment
+  - job:
```
mrtpain, line 106: We can try it but tbh I'm not a big fan of load or integration tests in CI/CD pipelines due to added length of deployments. In emergencies we want to be fast, waiting 10-20 minutes for a deployment could be a problem.

mrtpain (approved):

azure-pipelines.yml
drmathias: Yeah I think you make a good point. Let's keep this single test, it will still be useful as a health check and I see it being useful when attempting to automate deployment slot swaps

PR #155, opened 2021-12-20
[PAPI-304] Send maximum CRS amount when SwapCrsForExactSrc

drmathias:
Resolves slippage bug for quoted transaction


mrtpain (approved):
nice 👍

PR #156, opened 2021-12-20
[PAPI-307] Token validation improvements

drmathias:
* Amends OpenAPI definition so CRS is valid parameter for token parameter on `GET/POST /wallet/{address}/balance/{token}`.
* Force returning 401 if CRS is provided to `GET /wallet/{address}/allowance/{token}/spender/{spender}`, otherwise it returns 500 (call to full node fails)

### DIFF
--- a/src/Opdex.Platform.WebApi/Controllers/WalletsController.cs
+++ b/src/Opdex.Platform.WebApi/Controllers/WalletsController.cs
@@ -35,7 +35,7 @@ public class WalletsController : ControllerBase
     }
 
     /// <summary>Get Approved Allowance</summary>
-    /// <remarks>Retrieve the allowance of a spender for tokens owned by another wallet.</remarks>
+    /// <remarks>Retrieve the allowance of a spender for SRC tokens owned by another wallet.</remarks>
     /// <param name="address">Address of the wallet.</param>
     /// <param name="token">Address of the token.</param>
     /// <param name="spender">Address for the spender of the allowance.</param>
@@ -52,7 +52,7 @@ public class WalletsController : ControllerBase
                                                                                  [FromRoute] Address spender,
                                                                                  CancellationToken cancellationToken)
     {
-        if (token == Address.Cirrus) throw new InvalidDataException(nameof(token), "Token must be network address for an SRC token.");
+        if (token == Address.Cirrus) throw new InvalidDataException(nameof(token), "Address must be SRC token address.");
         var allowances = await _mediator.Send(new GetAddressAllowanceQuery(address, spender, token), cancellationToken);
         var response = _mapper.Map<ApprovedAllowanceResponseModel>(allowances);
         return Ok(response);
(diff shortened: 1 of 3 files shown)


src/Opdex.Platform.WebApi/OpenApi/Wallets/RefreshAddressBalanceOperationProcessor.cs
```diff
@@ -15,7 +15,7 @@ public bool Process(OperationProcessorContext context)
 
         var tokenAddressParameter = context.OperationDescription.Operation.Parameters.First(parameter => parameter.Name == "token");
         tokenAddressParameter.Schema.Example = "tF83sdXXt2nTkL7UyEYDVFMK4jTuYMbmR3";
-        tokenAddressParameter.Schema.MinLength = 30;
+        tokenAddressParameter.Schema.MinLength = 3;
         tokenAddressParameter.Schema.MaxLength = 42;
```
mrtpain, line 19: For this Refresh address balances can keep 30 min length I think, we won't be refreshing CRS balances unless we start to persist them.

src/Opdex.Platform.WebApi/OpenApi/Wallets/GetAddressBalanceByTokenOperationProcessor.cs
```diff
@@ -15,7 +15,7 @@ public bool Process(OperationProcessorContext context)
 
         var tokenAddressParameter = context.OperationDescription.Operation.Parameters.First(parameter => parameter.Name == "token");
         tokenAddressParameter.Schema.Example = "tF83sdXXt2nTkL7UyEYDVFMK4jTuYMbmR3";
-        tokenAddressParameter.Schema.MinLength = 30;
+        tokenAddressParameter.Schema.MinLength = 3;
```
mrtpain, line 18: This is interesting because the frontend has been working fine with this validation, wondering if it was ever working 🤔

src/Opdex.Platform.WebApi/Controllers/WalletsController.cs
```diff
@@ -100,9 +102,9 @@ public WalletsController(IMapper mapper, IMediator mediator)
     }
 
     /// <summary>Refresh Balance</summary>
-    /// <remarks>Retrieves and indexes the latest wallet public key balance for a token.</remarks>
-    /// <param name="address">The address of the wallet.</param>
-    /// <param name="token">The token to get the balance of.</param>
+    /// <remarks>Retrieves and indexes the latest balance of a token for an address.</remarks>
+    /// <param name="address">Address of the wallet.</param>
+    /// <param name="token">Address of the token to refresh the balance of, or CRS.</param>
```
mrtpain, line 107: Think we can drop CRS support from this endpoint and return 400

src/Opdex.Platform.WebApi/OpenApi/Wallets/GetAddressBalanceByTokenOperationProcessor.cs
drmathias: This doesn't perform validation, it is used to produce the OAS doc

src/Opdex.Platform.WebApi/Controllers/WalletsController.cs
drmathias: Sorted

src/Opdex.Platform.WebApi/OpenApi/Wallets/GetAddressBalanceByTokenOperationProcessor.cs
mrtpain: aaahhhh

mrtpain (approved):

PR #157, opened 2021-12-20
[PAPI-308] Mining pool route parameters

drmathias:
* Defines schema mining pool params in the OAS doc
* Tweak to token params so that endpoint responds with 400 when token cannot be CRS


src/Opdex.Platform.WebApi/OpenApi/OpenApiParameterExtensions.cs
```diff
@@ -0,0 +1,29 @@
+using NSwag;
+using Opdex.Platform.Common.Models;
+
+namespace Opdex.Platform.WebApi.OpenApi;
+
+public static class OpenApiParameterExtensions
```
mrtpain, line 6: Nice 😎

mrtpain (approved):

PR #159, opened 2021-12-22
[PAPI-312] Include zero liquidity flag

drmathias:
* Adds a flag for whether to include tokens with no liquidity
* Explores potential for returning average price data across markets for token endpoints

### DIFF
--- a/src/Opdex.Platform.Application/Assemblers/MarketTokenDtoAssembler.cs
+++ b/src/Opdex.Platform.Application/Assemblers/MarketTokenDtoAssembler.cs
@@ -24,6 +24,12 @@ public class MarketTokenDtoAssembler : IModelAssembler<MarketToken, MarketTokenD
     {
         var marketTokenDto = _mapper.Map<MarketTokenDto>(token);
 
+        if (marketTokenDto.Summary is null)
+        {
+            var summary = await _mediator.Send(new RetrieveTokenSummaryByMarketAndTokenIdQuery(token.Market.Id, token.Id, findOrThrow: false));
+            if (summary is not null) marketTokenDto.Summary = _mapper.Map<TokenSummaryDto>(summary);
+        }
+
         var liquidityPool = token.IsLpt
             ? await _mediator.Send(new RetrieveLiquidityPoolByAddressQuery(token.Address))
             : await _mediator.Send(new RetrieveLiquidityPoolBySrcTokenIdAndMarketIdQuery(token.Id, token.Market.Id));
--- a/src/Opdex.Platform.Application/Assemblers/MarketTokenDtoAssembler.cs
+++ b/src/Opdex.Platform.Application/Assemblers/MarketTokenDtoAssembler.cs
@@ -24,13 +24,6 @@ public class MarketTokenDtoAssembler : IModelAssembler<MarketToken, MarketTokenD
     {
         var marketTokenDto = _mapper.Map<MarketTokenDto>(token);
 
-        var summary = await _mediator.Send(new RetrieveTokenSummaryByMarketAndTokenIdQuery(token.Market.Id, token.Id, findOrThrow: false));
-
-        if (summary != null)
-        {
-            marketTokenDto.Summary = _mapper.Map<TokenSummaryDto>(summary);
-        }
-
         var liquidityPool = token.IsLpt
             ? await _mediator.Send(new RetrieveLiquidityPoolByAddressQuery(token.Address))
             : await _mediator.Send(new RetrieveLiquidityPoolBySrcTokenIdAndMarketIdQuery(token.Id, token.Market.Id));
--- a/src/Opdex.Platform.Application/EntryHandlers/Addresses/Mining/GetMiningPositionsWithFilterQueryHandler.cs
+++ b/src/Opdex.Platform.Application/EntryHandlers/Addresses/Mining/GetMiningPositionsWithFilterQueryHandler.cs
@@ -1,5 +1,5 @@
 using MediatR;
-using Opdex.Platform.Application.Abstractions.EntryQueries.Addresses;
+using Microsoft.Extensions.Logging;
 using Opdex.Platform.Application.Abstractions.EntryQueries.Addresses.Mining;
 using Opdex.Platform.Application.Abstractions.Models.Addresses;
 using Opdex.Platform.Application.Abstractions.Queries.Addresses.Mining;
@@ -16,11 +16,14 @@ public class GetMiningPositionsWithFilterQueryHandler : EntryFilterQueryHandler<
 {
     private readonly IMediator _mediator;
     private readonly IModelAssembler<AddressMining, MiningPositionDto> _miningPositionAssembler;
+    private readonly ILogger<GetMiningPositionsWithFilterQueryHandler> _logger;
 
-    public GetMiningPositionsWithFilterQueryHandler(IMediator mediator, IModelAssembler<AddressMining, MiningPositionDto> miningPositionAssembler)
+    public GetMiningPositionsWithFilterQueryHandler(IMediator mediator, IModelAssembler<AddressMining, MiningPositionDto> miningPositionAssembler, ILogger<GetMiningPositionsWithFilterQueryHandler> logger)
+        : base(logger)
     {
         _mediator = mediator ?? throw new ArgumentNullException(nameof(mediator));
         _miningPositionAssembler = miningPositionAssembler ?? throw new ArgumentNullException(nameof(miningPositionAssembler));
+        _logger = logger ?? throw new ArgumentNullException(nameof(logger));
     }
 
     public override async Task<MiningPositionsDto> Handle(GetMiningPositionsWithFilterQuery request, CancellationToken cancellationToken)
@@ -29,9 +32,15 @@ public class GetMiningPositionsWithFilterQueryHandler : EntryFilterQueryHandler<
 
         var positionsResults = positions.ToList();
 
+        _logger.LogTrace("Retrieved queried mining positions");
+
         var cursorDto = BuildCursorDto(positionsResults, request.Cursor, pointerSelector: result => result.Id);
 
+        _logger.LogTrace("Returning {ResultCount} results", positionsResults.Count);
+
         var dtos = await Task.WhenAll(positionsResults.Select(position => _miningPositionAssembler.Assemble(position)));
+
+        _logger.LogTrace("Assembled results");
 
         return new MiningPositionsDto { Positions = dtos, Cursor = cursorDto };
     }
--- a/src/Opdex.Platform.Application/EntryHandlers/Addresses/Mining/GetMiningPositionsWithFilterQueryHandler.cs
+++ b/src/Opdex.Platform.Application/EntryHandlers/Addresses/Mining/GetMiningPositionsWithFilterQueryHandler.cs
@@ -16,14 +16,12 @@ public class GetMiningPositionsWithFilterQueryHandler : EntryFilterQueryHandler<
 {
     private readonly IMediator _mediator;
     private readonly IModelAssembler<AddressMining, MiningPositionDto> _miningPositionAssembler;
-    private readonly ILogger<GetMiningPositionsWithFilterQueryHandler> _logger;
 
     public GetMiningPositionsWithFilterQueryHandler(IMediator mediator, IModelAssembler<AddressMining, MiningPositionDto> miningPositionAssembler, ILogger<GetMiningPositionsWithFilterQueryHandler> logger)
         : base(logger)
     {
         _mediator = mediator ?? throw new ArgumentNullException(nameof(mediator));
         _miningPositionAssembler = miningPositionAssembler ?? throw new ArgumentNullException(nameof(miningPositionAssembler));
-        _logger = logger ?? throw new ArgumentNullException(nameof(logger));
     }
 
     public override async Task<MiningPositionsDto> Handle(GetMiningPositionsWithFilterQuery request, CancellationToken cancellationToken)
@@ -32,15 +30,9 @@ public class GetMiningPositionsWithFilterQueryHandler : EntryFilterQueryHandler<
 
         var positionsResults = positions.ToList();
 
-        _logger.LogTrace("Retrieved queried mining positions");
-
         var cursorDto = BuildCursorDto(positionsResults, request.Cursor, pointerSelector: result => result.Id);
 
-        _logger.LogTrace("Returning {ResultCount} results", positionsResults.Count);
-
         var dtos = await Task.WhenAll(positionsResults.Select(position => _miningPositionAssembler.Assemble(position)));
-
-        _logger.LogTrace("Assembled results");
 
         return new MiningPositionsDto { Positions = dtos, Cursor = cursorDto };
     }
--- a/src/Opdex.Platform.Domain/Models/Tokens/TokenBase.cs
+++ b/src/Opdex.Platform.Domain/Models/Tokens/TokenBase.cs
@@ -46,7 +46,7 @@ public abstract class TokenBase : BlockAudit
     }
 
     protected TokenBase(ulong id, Address address, bool isLpt, string name, string symbol, int decimals, ulong sats, UInt256 totalSupply,
-                        ulong createdBlock, ulong modifiedBlock) : base(createdBlock, modifiedBlock)
+                        TokenSummary tokenSummary, ulong createdBlock, ulong modifiedBlock) : base(createdBlock, modifiedBlock)
     {
         Id = id;
         Address = address;
@@ -56,6 +56,7 @@ public abstract class TokenBase : BlockAudit
         Decimals = decimals;
         Sats = sats;
         TotalSupply = totalSupply;
+        TokenSummary = tokenSummary;
     }
 
     public ulong Id { get; }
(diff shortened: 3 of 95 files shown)


mrtpain:
This still a WIP or ready for review?

src/Opdex.Platform.Application/EntryHandlers/Addresses/Mining/GetMiningPositionsWithFilterQueryHandler.cs
```diff
@@ -29,10 +32,16 @@ public override async Task<MiningPositionsDto> Handle(GetMiningPositionsWithFilt
 
         var positionsResults = positions.ToList();
 
+        _logger.LogTrace("Retrieved queried mining positions");
+
         var cursorDto = BuildCursorDto(positionsResults, request.Cursor, pointerSelector: result => result.Id);
 
+        _logger.LogTrace("Returning {ResultCount} results", positionsResults.Count);
+
```
mrtpain, line 40: First thoughts, this is going to be quite a bit of logs, not sure what level we need to log at but it could be quite a bit.

Might be more efficient to wrap handlers within a timer and write logs with "starting" and "finished in ##ms" rather than step by step logs.

drmathias: A timer doesn't really add anything as time is included for all logs. It's good to be generous with trace logs, as they're the lowest granular level. We only ever need to turn them on when we want to trace execution time or debug something very granular. The idea behind putting in these logs is that we can see when database results get returned, when the cursor is built and when assembling the dtos happens. This will hopefully give us an idea which of these 3 steps might be the cause of performance concerns. Only logging at the start and end of handler execution won't provide this sort of insight.

mrtpain: I hear ya but...this will be **a lot** of logs and quite a bit of effort to go back and implement everywhere. 

It might be beneficial to take a simple route first, then come back and get granular if/where necessary.

We could create a new BaseHandler, with protected `StartTrace` and `StopTrace` methods and easily call those base methods passing in the name of the parent handler at the start and end of `Handle()`. Then we get the base info we need for a very minimal amount of work.

``` C#
// Imperfect example
public override async Task<MiningPositionsDto> Handle(GetMiningPositionsWithFilterQuery request, CancellationToken cancellationToken) 
{
    StartTrace();
    
    // Our actual handler logic
    
    StopTrace();
}

// Base Class Methods
protected void StartTrace() {
    _logger.LogTrace("{0}: Starting handler execution", _handlerName);
    _timer = new Stopwatch();
    _timer.Start();
}

protected void StopTrace() 
{
    _timer.Stop();
    _logger.LogTrace("{0}: Time elapsed: {1:hh\\:mm\\:ss}", _handlerName, stopwatch.Elapsed);
}
```

drmathias: I don't disagree with logging around handlers, it'll produce a lot of useful info.

This level of granularity is necessary due to the result of the load test. I'm not sure whether by *a lot* you're suggesting it's too granular? We won't have trace logs turned on unless we need them, really I don't see the downside of having a lot of logs when all of them have a purpose. 

```csharp
...
// gives us an idea of how long the database query takes
_logger.LogTrace("Retrieved queried address balances");

 var cursorDto = BuildCursorDto(balancesResults, request.Cursor, pointerSelector: result => result.Id);

 // tells us how long building the cursor takes as well as how many dtos will be assembled
 _logger.LogTrace("Returning {ResultCount} results", balancesResults.Count);

 var dtos = await Task.WhenAll(balancesResults.Select(balance => _assembler.Assemble(balance)));

 // lets us know how long assembling of the dtos takes
 _logger.LogTrace("Assembled results");
```

In fact, maybe removing the `Retrieved queried .........` logs and logging more granularly within the select handlers or `DbContext` instead would be more beneficial, so we can see a more precise result for query time.

src/Opdex.Platform.Infrastructure/Data/Handlers/Tokens/SelectTokensWithFilterQueryHandler.cs
```diff
@@ -64,18 +64,19 @@ public async Task<IEnumerable<Token>> Handle(SelectTokensWithFilterQuery request
...
 
     private static string QueryBuilder(SelectTokensWithFilterQuery request)
     {
-        var whereFilter = request.MarketId > 0
-            ? $" WHERE ts.{nameof(TokenSummaryEntity.MarketId)} = @{nameof(SqlParams.MarketId)}"
-            : $" WHERE (ts.{nameof(TokenSummaryEntity.MarketId)} IS NULL OR ts.{nameof(TokenSummaryEntity.MarketId)} = 0)";
+        var filterOnMarket = request.MarketId > 0;
+
+        var whereFilterBuilder = new StringBuilder();
+        if (filterOnMarket) whereFilterBuilder.Append($" WHERE ts.{nameof(TokenSummaryEntity.MarketId)} = @{nameof(SqlParams.MarketId)}");
```
mrtpain, line 79: I think we have to keep the ternary logic and the initial join. Multiple markets can return multiple rows because of the join on `token_summary`. So if we filter for a specific market, it'll be fine, if not, then it can return the same token for each market its within.

Maybe its time to split this out into a MarketTokens vs Tokens filter

src/Opdex.Platform.Infrastructure/Data/Handlers/Tokens/SelectTokensWithFilterQueryHandler.cs
```diff
@@ -35,14 +36,13 @@ public class SelectTokensWithFilterQueryHandler : IRequestHandler<SelectTokensWi
                 t.{nameof(TokenEntity.TotalSupply)},
                 t.{nameof(TokenEntity.CreatedBlock)},
                 t.{nameof(TokenEntity.ModifiedBlock)},
-                ts.{nameof(TokenSummaryEntity.PriceUsd)},
-                ts.{nameof(TokenSummaryEntity.DailyPriceChangePercent)}
+                AVG(ts.{nameof(TokenSummaryEntity.PriceUsd)}) AS {nameof(TokenSummaryEntity.PriceUsd)},
+                AVG(ts.{nameof(TokenSummaryEntity.DailyPriceChangePercent)}) AS {nameof(TokenSummaryEntity.DailyPriceChangePercent)}
```
mrtpain, line 40: This is going to average all of the results prices, from all tokens.

it will require either a new `SELECT` internally here or another query altogether to specifically average a one token across markets. 

I think we should just return the price based on the specific market, for market tokens.

We should probably split these queries out for Market/Global tokens and not allow filtering based on averaged token summary data for global tokens.

src/Opdex.Platform.Application/EntryHandlers/Addresses/Mining/GetMiningPositionsWithFilterQueryHandler.cs
mrtpain: It's a code smell having this many logs around and it's a bit tedious to continue doing. 

We know a total requests time and we can see the full flow of a request through the traceId. Logging these only in the infrastructure layer is probably better, maybe the Cirrus Modules before the go out to the Node, CMC modules going to their API, and DB context.

If we do this, theres only a few places that need this logging and we could add a DB context optional param at the end for the `operationName` so we'll know exactly which query it was. We'd just add `nameof(OurQueryHandler)`.

it might be also worth using a StopWatch and logging the elapsed time within these logs rather than us looking at DateTimes within logs. 

None of our internal operations should be long running I don't think without needing the help of this external information.

src/Opdex.Platform.Infrastructure/Data/Handlers/Tokens/SelectTokensWithFilterQueryHandler.cs
drmathias: I can do some testing around this, my belief is that the `GROUP BY` will limit it to return only a single row for each token even with multiple markets.

src/Opdex.Platform.Infrastructure/Data/Handlers/Tokens/SelectTokensWithFilterQueryHandler.cs
drmathias: The `GROUP BY t.{nameof(TokenEntity.Id)}` should cause the average to be per token

src/Opdex.Platform.Application/EntryHandlers/Addresses/Mining/GetMiningPositionsWithFilterQueryHandler.cs
drmathias: I disagree about it being a code smell, but I'll remove the logs within these handlers and happy to replace with logging around the external calls like you suggest. That should allow us to gather a good amount of information to work with.

src/Opdex.Platform.Infrastructure/Data/Handlers/Tokens/SelectTokensWithFilterQueryHandler.cs
drmathias: I appear to have got it working by defining a one to one relationship using dapper multi-mapping. Taking the following data:
image

Call to `GET /tokens` returns
```json
{
    "address": "PRpQxG3nFPzs4KZfKgFYteYkoBHGQKUH1V",
    "name": "Bitcoin (Wrapped)",
    "symbol": "xBTC",
    "decimals": 8,
    "sats": 100000000,
    "totalSupply": "21000000.00000000",
    "summary": {
        "priceUsd": 528.411446000000,
        "dailyPriceChangePercent": -1.430000000000,
        "modifiedBlock": 1472
    }
}
```

Call to `GET /markets/PQKXghAygiEGfE7TaLymHKGiZ42nG7DvrM/tokens` returns
```json
{
    "market": "PQKXghAygiEGfE7TaLymHKGiZ42nG7DvrM",
    "liquidityPool": "PKQc4Lj7pTy9se1B3qBLDQv33nZ34bnSP1",
    "address": "PRpQxG3nFPzs4KZfKgFYteYkoBHGQKUH1V",
    "name": "Bitcoin (Wrapped)",
    "symbol": "xBTC",
    "decimals": 8,
    "sats": 100000000,
    "totalSupply": "21000000.00000000",
    "summary": {
        "priceUsd": 536.290000000000,
        "dailyPriceChangePercent": 1.430000000000,
        "modifiedBlock": 1428
    }
}
```

Splitting them into two handlers can still be done, though probably isn't necessary unless there's a strong reason to optimise them.

src/Opdex.Platform.Application/Assemblers/MarketTokenDtoAssembler.cs
```diff
@@ -24,13 +24,6 @@ public async Task<MarketTokenDto> Assemble(MarketToken token)
     {
         var marketTokenDto = _mapper.Map<MarketTokenDto>(token);
 
-        var summary = await _mediator.Send(new RetrieveTokenSummaryByMarketAndTokenIdQuery(token.Market.Id, token.Id, findOrThrow: false));
-
-        if (summary != null)
-        {
-            marketTokenDto.Summary = _mapper.Map<TokenSummaryDto>(summary);
-        }
-
```
mrtpain, line 33: I think this is still necessary for getting singular market tokens. so maybe wrap in `if(marketTokenDto.Summary == null)`

src/Opdex.Platform.Domain/Models/Tokens/TokenBase.cs
```diff
@@ -66,4 +67,5 @@ protected TokenBase(Address address, bool isLpt, string name, string symbol, int
     public int Decimals { get; }
     public ulong Sats { get; }
     public UInt256 TotalSupply { get; protected set; }
-}
\ No newline at end of file
+    public TokenSummary TokenSummary { get; }
```
mrtpain, line 70: nit - `Summary` is cleaner since we're within the Token already

src/Opdex.Platform.Infrastructure.Abstractions/Data/Models/Tokens/TokenEntity.cs
```diff
@@ -13,4 +13,5 @@ public class TokenEntity : AuditEntity
     public byte Decimals { get; set; }
     public ulong Sats { get; set; }
     public UInt256 TotalSupply { get; set; }
-}
\ No newline at end of file
+    public TokenSummaryEntity TokenSummary { get; set; }
```
mrtpain, line 16: nit - `Summary` instead though it doesn't matter much way down here

src/Opdex.Platform.Infrastructure/Data/DbContext.cs
```diff
@@ -30,7 +30,7 @@ public async Task<IEnumerable<TEntity>> ExecuteQueryAsync<TEntity>(DatabaseQuery
 
     public async Task<IEnumerable<TReturn>> ExecuteQueryAsync<TFirst, TSecond, TReturn>(DatabaseQuery query,
                                                                                         Func<TFirst, TSecond, TReturn> map,
-                                                                                        string splitOn)
+                                                                                        string splitOn = "Id")
```
mrtpain, line 33: Another nit, just because we have this done different elsewhere and this can be an easy area to miss causing a bugs in my past experience. 

I think we should remove the default "Id" from here and the interface and instead follow how its done in this query:
https://github.com/Opdex/opdex-v1-api/blob/main/src/Opdex.Platform.Infrastructure/Data/Handlers/Transactions/SelectTransactionsForSnapshotRewindQueryHandler.cs

src/Opdex.Platform.Infrastructure/Data/Handlers/Tokens/SelectTokensWithFilterQueryHandler.cs
mrtpain: Ok thats nice if it works! Only suggestion is to wrap in `ROUND(AVG(...), 8 )` so our percent and usd decimals stick to 8 decimals.

src/Opdex.Platform.Infrastructure/PlatformInfrastructureMapperProfile.cs
```diff
@@ -426,7 +426,7 @@ public PlatformInfrastructureMapperProfile()
             .ForMember(dest => dest.Symbol, opt => opt.MapFrom(src => src.Symbol))
             .ForMember(dest => dest.Decimals, opt => opt.MapFrom(src => src.Decimals))
             .ForMember(dest => dest.Sats, opt => opt.MapFrom(src => src.Sats))
-            .ForMember(dest => dest.TotalSupply, opt => opt.MapFrom(src => src.TotalSupply))
+            .ForMember(dest => dest.TokenSummary, opt => opt.MapFrom(src => src.TokenSummary))
             .ForMember(dest => dest.CreatedBlock, opt => opt.MapFrom(src => src.CreatedBlock))
```
mrtpain, line 430: This we want to keep total supply mapping here

mrtpain:
looks good, few comments. I like the `modifiedBlock` on the token summary, I think I'll create a story to return that and maybe createdBlock also.

src/Opdex.Platform.Application/Assemblers/MarketTokenDtoAssembler.cs
drmathias: Good point, I'd missed this. Deciding whether or not to add back this call or get the token summary within the query. Potentially we have a flag on the select token query for whether to load related entities (the token summary), then we avoid any performance penalties where the summary isn't needed. Although for now will keep it simple and add back the retrieval call in the assembler

src/Opdex.Platform.Infrastructure/PlatformInfrastructureMapperProfile.cs
drmathias: Re-added

src/Opdex.Platform.Infrastructure/Data/Handlers/Tokens/SelectTokensWithFilterQueryHandler.cs
drmathias: Added ROUND

src/Opdex.Platform.Infrastructure/Data/DbContext.cs
drmathias: Nice, I've followed how you've done it there, so the mapping is done within the split func rather than the summary being defined on the entity.

src/Opdex.Platform.Application/Assemblers/MarketTokenDtoAssembler.cs
drmathias: Thinking retrieving the summary from within the select handler is better. Since now we have `GET /markets/{market}/tokens{token}`, `GET /markets/{market}/tokens` and `GET /tokens` returning the summary, yet `GET/tokens/{token}` is not. For consistency and since we can, maybe we should.

mrtpain (approved):

mrtpain:
this can be merged but we can't swap slots yet for devnet so it won't really be live until then.

PR #161, opened 2021-12-23
[PAPI-313] Index status rewind status fix

drmathias:


src/Opdex.Platform.WebApi/Controllers/IndexController.cs
```diff
@@ -99,21 +97,28 @@ public async Task<IActionResult> ResyncFromDeployment(ResyncFromDeploymentReques
...
     [Authorize(Policy = "AdminOnly")]
     [ProducesResponseType(StatusCodes.Status204NoContent)]
     [ProducesResponseType(typeof(ProblemDetails), StatusCodes.Status401Unauthorized)]
     [ProducesResponseType(typeof(ProblemDetails), StatusCodes.Status403Forbidden)]
-    public async Task<ActionResult> Rewind(RewindRequest request)
+    public async Task<ActionResult> Rewind(RewindRequest request, CancellationToken cancellationToken)
     {
-        var tryLock = await _mediator.Send(new MakeIndexerLockCommand(IndexLockReason.Rewind), CancellationToken.None);
+        var latestSyncedBlock = await _mediator.Send(new RetrieveLatestBlockQuery(), cancellationToken);
+        if (latestSyncedBlock.Height <= request.Block) throw new InvalidDataException("block", "Block height must be less than current synced height.");
```
mrtpain, line 110: This can be removed, we throw invalid data exception within `CreateRewindToBlock` command when we attempt this same logic. There are a couple locations that call that method, so I think the check fits within there rather than moving it out.

src/Opdex.Platform.WebApi/Controllers/IndexController.cs
```diff
@@ -99,21 +97,28 @@ public async Task<IActionResult> ResyncFromDeployment(ResyncFromDeploymentReques
...
         if (!tryLock) throw new IndexingAlreadyRunningException();
 
         try
         {
+            var blockHash = await _mediator.Send(new RetrieveCirrusBlockHashByHeightQuery(request.Block), cancellationToken);
+            var rewindBlock = await _mediator.Send(new RetrieveCirrusBlockReceiptByHashQuery(blockHash), cancellationToken);
             var rewound = await _mediator.Send(new CreateRewindToBlockCommand(request.Block), CancellationToken.None);
-            if (!rewound) throw new Exception("Indexer rewind unexpectedly failed.");
+            if (!rewound) throw new Exception($"Failure rewinding database to block height: {request.Block}");
+            await _mediator.Send(new ProcessLatestBlocksCommand(rewindBlock, _network), CancellationToken.None);
```
mrtpain, line 121: - `blockHash` can be retrieved by height from the database using `request.Block` probably faster than asking the FN 
- Retrieving `blockHash` and `rewindBlock` can be moved below `if (!rewound) ...` since they aren't used yet and may not need to be if rewinding fails

src/Opdex.Platform.Domain/Models/IndexLock.cs
```diff
@@ -22,5 +22,5 @@ public IndexLock(bool available, bool locked, string instanceId, IndexLockReason
 
 public enum IndexLockReason
 {
-    Rewind = 1, Index = 2, Deploy = 3
+    Deploying = 1, Indexing = 2, Rewinding = 3
```
mrtpain, line 25: Not to myself, will need to update client for maintenance notifications with this pr

mrtpain:
Couple minor comments but for the most part lgtm. Will need client fixes when deployed. I'll test it out locally early next week too prior to merging. Nice work 👍 the added benefit of knowing that we're indexing because we're actually rewinding will be nice to know.

src/Opdex.Platform.WebApi/Controllers/IndexController.cs
mrtpain: Took care of the first two, still thinking on the third.

mrtpain (approved):

src/Opdex.Platform.WebApi/Controllers/IndexController.cs
drmathias: Thanks for taking care of it. I see the `ProcessLatestBlocks` is still there, did you decide your 3rd point wasn't an issue? I don't see that there's a timeout there

PR #166, opened 2022-01-10
[PAPI-321] Remove mempool check

drmathias:
Removes the mempool check in favour of checking the database for an indexed transaction


drmathias:
The unit test to test backoffs takes ~15 seconds. Not sure if dotnet test is running in parallel by default, though we can try to update our CI to specify `dotnet test -- xunit.parallelizeAssembly=true` if test runs are taking too long

mrtpain (approved):

PR #167, opened 2022-01-11
[PAPI-324] CoinGecko price feed

drmathias:


src/Opdex.Platform.Infrastructure/PlatformInfrastructureServiceCollectionExtensions.cs
```diff
@@ -384,6 +388,15 @@ private static void AddCirrusServices(IServiceCollection services, CirrusConfigu
         services.AddTransient<IRequestHandler<CallCirrusCreateSmartContractCommand, Sha256>, CallCirrusCreateSmartContractCommandHandler>();
     }
 
+    private static void AddCoinGeckoServices(IServiceCollection services)
+    {
+        services.AddTransient<ICoinGeckoClient, CoinGeckoClient>(_ => CoinGeckoClient.Instance);
```
mrtpain, line 393: I'd have to check but maybe you know, does this have a retry policy? 

We'd also probably want to include a circuit breaker policy too. Circuit breakers after a # of 500 errors will throw exceptions we can alert on, back off for # seconds, then open back up again.

src/Opdex.Platform.Infrastructure/Clients/CoinGeckoApi/Handlers/CallCoinGeckoGetStraxLatestPriceQueryHandler.cs
```diff
@@ -0,0 +1,32 @@
...
+        _client = client;
+        _logger = logger;
+    }
+
+    public async Task<decimal> Handle(CallCoinGeckoGetStraxLatestPriceQuery request, CancellationToken cancellationToken)
+    {
+        var result = await _client.CoinsClient.GetAllCoinDataWithId(CoinGeckoTokens.STRAX, "false", false, true, false, false, false);
+        if ((result?.MarketData?.CurrentPrice?.TryGetValue(CoinGeckoTokens.USD, out var usdPrice) ?? false) && usdPrice.HasValue)
+        {
+            return result.MarketData.CurrentPrice[CoinGeckoTokens.USD]!.Value;
```
mrtpain, line 26: Can we just return `usdPrice` here?

mrtpain (approved):
Retry policy and circuit breaker comment are probably worth looking into, otherwise lgtm as usual 👍

src/Opdex.Platform.Infrastructure/PlatformInfrastructureServiceCollectionExtensions.cs
drmathias: Doesn't look to be possible as is, I've forked and amended it so it's possible to configure policies. Opened a PR  https://github.com/tosunthex/CoinGecko/pull/37#issue-1100299073

src/Opdex.Platform.Infrastructure/Clients/CoinGeckoApi/Handlers/CallCoinGeckoGetStraxLatestPriceQueryHandler.cs
drmathias: Not sure what I was thinking here 😅, changed

src/Opdex.Platform.Infrastructure/PlatformInfrastructureServiceCollectionExtensions.cs
mrtpain: ok this is fine for now then. When we come back to this for price implementation, we may use our own client vs their package then until they support it. Retry and Circuit breakers are pretty important for resiliency

PR #168, opened 2022-01-11
[PAPI-322] Persist Failed Transaction Errors

drmathias:
* Persists raw error to transaction table
* Returns out both friendly and raw error
* Additional fix to prevent returning out empty addresses out as empty string in API responses

Tested locally returning 50 failed transactions from `GET /transactions`, performance seems completely fine


src/Opdex.Platform.WebApi/Mappers/TransactionErrorProcessor.cs
```diff
@@ -49,7 +48,7 @@ private string ProcessError(string rawError, TryParseFriendlyErrorMessage tryPar
 
         using (_logger.BeginScope(errorProperties))
         {
-            _logger.LogWarning("Transaction error occurred.");
+            _logger.LogWarning("Transaction error occurred");
         }
```
mrtpain, line 52: I think we can cut these from logs and alerts. Generally transaction errors are out of our control and can occur for many reasons, not necessarily at all related to the api.

I think in the future, if we wanted, we can create an alert on # of failed transactions within a time limit and % threshold, alerting us if we're seeing excessive errors over time, but even then I think our _only_ fault could be bad nodes, in which other alerts should go off prior.

what do you think?

src/Opdex.Platform.WebApi/Models/Responses/Transactions/TransactionErrorResponseModel.cs
```diff
@@ -0,0 +1,7 @@
+namespace Opdex.Platform.WebApi.Models.Responses.Transactions;
+
+public class TransactionErrorResponseModel
+{
+    public string Friendly { get; set; }
+    public string Raw { get; set; }
+}
```
mrtpain, line 7: nit could add some comments

src/Opdex.Platform.WebApi/Startup.cs
```diff
@@ -126,7 +126,8 @@ public void ConfigureServices(IServiceCollection services)
 
         services.AddHttpClient();
 
-        // Automapper Profiles
+        // Automapper
+        services.AddTransient<TransactionErrorMappingAction>();
```
mrtpain, line 130: This is cool

mrtpain (approved):

src/Opdex.Platform.WebApi/Mappers/TransactionErrorProcessor.cs
drmathias: Currently all transactions picked up by the indexer hit this `TransactionErrorProcessor`, causing these transaction errors to be logged. These changes mean this is no longer the case, it'll only occur on persisted transactions which should get filtered to only contain Opdex transactions. Although logging for transaction errors isn't really helpful when we're persisting them and I agree it needs some changes. I'll make some amendments so we're only logging failed Opdex transactions that cannot be parsed.

drmathias:
Quickly covered off PAPI-326 in this PR as well

mrtpain (approved):

PR #170, opened 2022-01-14
[PAPI-319] Market token request validation

drmathias:
Resolves bug where swaps with zero amounts were returning a 500 Internal Server Error, instead they will return 400 Bad Request.

### DIFF
--- a/src/Opdex.Platform.WebApi/Validation/MarketTokens/SwapRequestValidator.cs
+++ b/src/Opdex.Platform.WebApi/Validation/MarketTokens/SwapRequestValidator.cs
@@ -1,0 +1,34 @@
+using FluentValidation;
+using Opdex.Platform.Common.Models;
+using Opdex.Platform.WebApi.Models.Requests.MarketTokens;
+
+namespace Opdex.Platform.WebApi.Validation.MarketTokens;
+
+public class SwapRequestValidator : AbstractValidator<SwapRequest>
+{
+    public SwapRequestValidator()
+    {
+        RuleFor(r => r.TokenOut)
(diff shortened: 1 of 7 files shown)


src/Opdex.Platform.WebApi/Validation/MarketTokens/SwapRequestValidator.cs
```diff
@@ -0,0 +1,34 @@
...
+            .GreaterThanOrEqualTo(req => req.TokenInAmount).WithMessage("Token in maximum amount cannot be less than token in amount.");
+        When(r => r.TokenInExactAmount, () =>
+        {
+            RuleFor(r => r.TokenInMaximumAmount).Must((req, tokenInMaximumAmount) => tokenInMaximumAmount == req.TokenInAmount)
+                                                .WithMessage("Token in maximum amount must equal token in amount, since exact amount is specified.");
+        });
+        RuleFor(r => r.TokenOutMinimumAmount)
+            .MustBeValidSrcValue().WithMessage("Token out minimum amount must contain 18 decimal places or less.")
+            .GreaterThan(FixedDecimal.Zero).WithMessage("Token out minimum amount must be greater than zero.")
+            .LessThanOrEqualTo(req => req.TokenOutAmount).WithMessage("Token out minimum amount cannot be greater than token out amount.");
```
mrtpain, line 30: We might want another `When` for `!r.TokenInExactAmount` that validates that `TokenOutMinimumAmount == TokenOutAmount`

mrtpain (approved):
Only 1 suggestion, lgtm 👍

PR #174, opened 2022-01-18
[PAPI-298] Refresh Low Volume Snapshots

mrtpain:
Work in progress PR to refresh USD prices for liquidity pool and token snapshots that have had no volume within a set block threshold. 

Forces USD prices to update after no activity to ensure that low volume tokens and liquidity pools have refreshed USD pricing within the last 15 minutes. Otherwise, pricing is updated after successful transactions in real time.


src/Opdex.Platform.Application/EntryHandlers/LiquidityPools/Snapshots/ProcessStaleLiquidityPoolSnapshotsCommandHandler.cs
```diff
@@ -0,0 +1,77 @@
...
+
+                // If its not null, then its a stale snapshot, refresh it first
+                if (liquidityPool != null)
+                {
+                    var stakingToken = await _mediator.Send(new RetrieveTokenByIdQuery(market.StakingTokenId), CancellationToken.None);
+                    var lpToken = await _mediator.Send(new RetrieveTokenByAddressQuery(liquidityPool.Address), CancellationToken.None);
+
+                    foreach (var snapshotType in snapshotTypes)
+                    {
+                        await _mediator.Send(new ProcessDailyLiquidityPoolSnapshotRefreshCommand(liquidityPool.Id, liquidityPool.MarketId, stakingToken,
```
drmathias, line 49: I'm a little confused by the name of the this command. `ProcessDaily` implies it's for the daily interval although we're passing Hourly and Daily as snapshot types

src/Opdex.Platform.Application/EntryHandlers/LiquidityPools/Snapshots/ProcessStaleLiquidityPoolSnapshotsCommandHandler.cs
```diff
@@ -0,0 +1,77 @@
...
+                                                                                                 request.BlockHeight));
+                    }
+                }
+
+                var stakingTokenSnapshot = await _mediator.Send(new RetrieveTokenSnapshotWithFilterQuery(market.StakingTokenId, market.Id, request.BlockTime,
+                                                                                                         SnapshotType.Hourly), CancellationToken.None);
+                stakingTokenUsd = stakingTokenSnapshot.Price.Close;
+            }
+
+            foreach (var liquidityPool in marketGroup.Where(pool => pool.SrcTokenId != market.StakingTokenId))
```
drmathias, line 60: Extracting the loop body to a task and using `Task.WhenAll` may be beneficial here if we have a lot of liquidity pools. Or using `Parallel.ForEachAsync` may be even better, although I think we'd need to benchmark to see what's best. `Parallel.ForEach` used to be discouraged for IO operations, though I'm unsure if that's still the case with the async method introduced in .NET 6

drmathias (approved):

src/Opdex.Platform.Application/EntryHandlers/LiquidityPools/Snapshots/ProcessStaleLiquidityPoolSnapshotsCommandHandler.cs
mrtpain: I want to read a bit more to get a better understanding but you may be right, `ForEachAsync` may be better

PR #177, opened 2022-01-19
[PAPI-103] Api versioning

drmathias:
* Requires prepending version to all API routes, so for example we can call `/v1/status`
* Returns `api-supported-versions` header for all API calls, so consumers can see they can potentially upgrade
* Returns 400 with error message if an unsupported version is used


mrtpain:
Wondering if we should include `v1` in the opdex configuration ApiUrl, or if we should add `v1/...` to Transaction callback endpoint and auth callback endpoints in configuration.

Will need to apply changes to the frontend to match this. 

Also wondering if we should add `v1` to the `socket` hub.

---

Just those things to figure out and implement along with conflicts.

drmathias:
> Wondering if we should include `v1` in the opdex configuration ApiUrl, or if we should add `v1/...` to Transaction callback endpoint and auth callback endpoints in configuration.

Probably individual is more flexible, although I can't see a need to update these separately. Can stick with the /v1 on the ApiUrl and then we can always change config without disruption, if the need arises

> Also wondering if we should add `v1` to the `socket` hub.

I think we should, just in case

PR #184, opened 2022-01-21
[PAPI-331] Notify Broadcasted Transaction Improvements

mrtpain:
- Checks for raw transaction change address
    - Attempt to find raw transaction, 5 retries, 3 seconds in between, not found after retries throws `NotFoundException`
    - If change address is found, validates the requests public key matches change address
        - On success, notify
        - On failure, don't notify
    - If change address is not found, uses the requests public key for notification

Ultimately a shorter term fix until we work into deserializing the raw transaction in API or via FN, to notify the _actual_ transaction sender.

### DIFF
--- a/src/Opdex.Platform.Application/EntryHandlers/Transactions/CreateNotifyUserOfTransactionBroadcastCommandHandler.cs
+++ b/src/Opdex.Platform.Application/EntryHandlers/Transactions/CreateNotifyUserOfTransactionBroadcastCommandHandler.cs
@@ -26,16 +26,16 @@ public class CreateNotifyUserOfTransactionBroadcastCommandHandler : IRequestHand
         var transaction = await _mediator.Send(new RetrieveTransactionByHashQuery(request.TransactionHash, false), cancellationToken);
         if (transaction is not null) return false;
 
-        // find transaction on node, this isn't _really_ the sender, is the change output address
-        var sender = await _mediator.Send(new RetrieveCirrusUnverifiedTransactionSenderByHashQuery(request.TransactionHash), cancellationToken);
+        // find transaction on node, this isn't _really_ the sender, is the change output address, can throw not found if the transaction is not found
+        var changeAddress = await _mediator.Send(new RetrieveCirrusUnverifiedTransactionSenderByHashQuery(request.TransactionHash), cancellationToken);
 
-        // If there is a found "sender" above, it must match the provided sender through the callback
-        if (sender != Address.Empty && sender != request.Sender) return false;
+        // If a change address is found and it doesn't match the alleged sender, don't notify
+        if (changeAddress != Address.Empty && changeAddress != request.Sender) return false;
 
-        // If a "sender" wasn't found, just return out the sender sent
-        sender = sender == Address.Empty ? request.Sender : sender;
+        // If a change address wasn't found, trust the requests sender, notify anyways
+        var senderToNotify = changeAddress == Address.Empty ? request.Sender : changeAddress;
 
-        await _mediator.Send(new MakeNotifyUserOfTransactionBroadcastCommand(sender, request.TransactionHash), cancellationToken);
+        await _mediator.Send(new MakeNotifyUserOfTransactionBroadcastCommand(senderToNotify, request.TransactionHash), cancellationToken);
         return true;
     }
 }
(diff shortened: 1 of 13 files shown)


src/Opdex.Platform.Application/EntryHandlers/Transactions/CreateNotifyUserOfTransactionBroadcastCommandHandler.cs
```diff
@@ -24,11 +24,16 @@ public async Task<bool> Handle(CreateNotifyUserOfTransactionBroadcastCommand req
...
-        if (sender == Address.Empty) return false;
+        // find transaction on node, this isn't _really_ the sender, is the change output address, can throw not found if the transaction is not found
+        var changeAddress = await _mediator.Send(new RetrieveCirrusUnverifiedTransactionSenderByHashQuery(request.TransactionHash), cancellationToken);
 
-        await _mediator.Send(new MakeNotifyUserOfTransactionBroadcastCommand(sender, request.TransactionHash), cancellationToken);
+        // If a change address is found and it doesn't match the alleged sender, don't notify
+        if (changeAddress != Address.Empty && changeAddress != request.Sender) return false;
+
+        // If a change address wasn't found, trust the requests sender, notify anyways
+        var senderToNotify = changeAddress == Address.Empty ? request.Sender : changeAddress;
```
drmathias, line 34: Surely this would allow someone to send notifications to a logged in user. Unlikely to be a realistic problem, at least for the time we have this temporary solution.

src/Opdex.Platform.WebApi/Models/Requests/Transactions/TransactionBroadcastNotificationRequest.cs
```diff
@@ -12,4 +12,6 @@ public class TransactionBroadcastNotificationRequest
     /// </summary>
     /// <example>402aa2241adb7b04d07d4dbc89f8aae72fa1c11f9bd2bd9013222cd774ed39fe</example>
     public Sha256 TransactionHash { get; set; }
+
+    public Address PublicKey { get; set; }
```
drmathias, line 16: This needs documenting in openapi.yml

nitpick: Consider renaming to Address, as I think public key is a more ambiguous term.

src/Opdex.Platform.WebApi/Validation/Transactions/TransactionBroadcastNotificationValidator.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Opdex.Platform.WebApi.Validation.Transactions;
+
+public class TransactionBroadcastNotificationValidator : AbstractValidator<TransactionBroadcastNotificationRequest>
+{
+    public TransactionBroadcastNotificationValidator()
+    {
+        RuleFor(request => request.TransactionHash)
+            .NotEmpty().WithMessage("Transaction hash must not be empty.")
+            .NotEqual(new Sha256()).WithMessage("Transaction hash must be valid.");
```
drmathias, line 13: Believe this can be removed. The Sha256 JSON converter will ensure it's not empty and `new Sha256()` is equal to `0000000000000000000000000000000000000000000000000000000000000000` which is theoretically a valid hash. Someone would have to provide that value explicitly because the JSON converter won't convert null/empty to this, it'll instead throw an exception which should get treated as a 400 Bad Request.

src/Opdex.Platform.WebApi/Models/Requests/Transactions/TransactionBroadcastNotificationRequest.cs
mrtpain: Renaming would require changing SSAS spec

src/Opdex.Platform.Application/EntryHandlers/Transactions/CreateNotifyUserOfTransactionBroadcastCommandHandler.cs
mrtpain: Yeah, we can't be 100% accurate right now. 
Transaction receipts from the full node only return for mined transactions. 
Raw transactions we can get from the mempool but can't deserialize everything atm to get the sender without further work either in our node repo, bringing in some stratis packages to here, or PR to the SFN repo.

drmathias (approved):

PR #186, opened 2022-01-23
[PAPI-292] Remove Guest JWT Requirement

mrtpain:
- Removes legacy auth endpoint
- Removes Authorize attribute from public endpoints
- Wallet JWT have a 24h expiration time (up from 1h)
- Adds auth to refresh token balance endpoint
- Removes Deploy controller


mrtpain:
cc @drmathias merging this now, you can review and comment and I'll follow up with another PR with anything. 

Tried to work alongside this but its frustrating between other branches and UI work w/ conflicts etc.

src/Opdex.Platform.WebApi/Controllers/AuthController.cs
```diff
@@ -96,37 +94,4 @@ public class AuthController : ControllerBase
 
         return NoContent();
     }
-
-    /// <summary>Authorize</summary>
-    /// <remarks>Authorizes access to a specific market</remarks>
-    /// <param name="wallet">The wallet public key of the user</param>
-    /// <returns>An access token</returns>
-    [HttpPost("authorize")]
-    public IActionResult Authorize([FromQuery] Address wallet)
```
drmathias, line 105: When developing locally, have you come up with an easy way to get the auth token? All I can think of with this now removed is to debug and set a breakpoint when the bearer token is built, then auth with the SASS endpoint

mrtpain: Good point, for devnet and up you can rip the bearer token from any api request in the network tab of dev tools. For local, that is not really possible without also running a local client. 

Maybe I put a separate devnet only endpoint back for this purpose only, local/devnet token retrieval

mrtpain: https://github.com/Opdex/opdex-v1-api/pull/191

PR #189, opened 2022-01-24
[PAPI-327] Filter options and enum definitions

drmathias:
* Multiple enum filters for certificates
* Amend token query params to use token attributes over token type
* Define enum `Any` options


drmathias:
Can hold off on review, realised I missed out the get vault proposals endpoint

mrtpain (approved):
lgtm, I'll merge it alongside the UI fixes

PR #190, opened 2022-01-24
[PAPI-336] Add request interceptor for Swagger URL

drmathias:
Relies on the server address amendment to `openapi.yml` present in https://github.com/Opdex/opdex-v1-api/pull/189


drmathias:
Closing, as replacing server will be more effective

mrtpain:
@drmathias 

Interceptor catches and changes request URL based on current swagger domain. 
JS injection changes the select menu, removing the default test-api selection. 

This is still not very ideal as we are picking things out based on versions. Really this is all so our docs can maintain a `v1-test-api` selection and details. 

We can alleviate all of this and remove the interceptor and JS files entirely if we removed `servers` form the OpenApi.yml file and include relative paths like `/v1`. However, this would point our ReadMe docs to mainnet if we release after mainnet deployments. 

The other option is to release after testnet deployments so readme should use the testnet domain, then we can also remove this work..._I think_

drmathias:
> This is still not very ideal as we are picking things out based on versions. Really this is all so our docs can maintain a `v1-test-api` selection and details.
> 
> We can alleviate all of this and remove the interceptor and JS files entirely if we removed `servers` form the OpenApi.yml file and include relative paths like `/v1`. However, this would point our ReadMe docs to mainnet if we release after mainnet deployments.
> 
> The other option is to release after testnet deployments so readme should use the testnet domain, then we can also remove this work..._I think_

Nice, I didn't realise we could use a relative URL. It seems so obvious 😆
That seems much better, and I agree we move the pipeline step back to run after testnet deployment

PR #191, opened 2022-01-25
[PAPI-292] Additional Auth Methods

mrtpain:
- Adds **GET** `/auth` endpoint to get a new StratisId to sign
- Migrated **POST** `/auth` to **POST** `/auth/callback` for default SSAS flows with sockets
- Adjusts **POST** `/auth` to validate signatures and return JWT directly

---

The additional endpoints provide a way for API based clients to authenticate themselves rather than their users such as within Postman, or a bot based API for example for retrieving transaction quotes and acting upon the results.

Naming probably could be better, the new auth routes could add `Direct` to the naming and the socket based flows `Socket` or `Callback` etc. or leave as is. More of a nit, but could clarify the differences more for clients.

### DIFF
--- a/docs/openapi.yml
+++ b/docs/openapi.yml
@@ -2767,13 +2767,33 @@ components:
           description: Whether the vote is in favor of the proposal
 paths:
   /auth:
+    get:
+      tags:
+        - Authentication
+      summary: Get Stratis Id
+      description: >
+        Returns a new Stratis Id to sign and authenticate.
+        See the [specification](https://github.com/Opdex/SSAS) for further detail.
+      operationId: getStratisId
+      responses:
+        200:
+          description: Signature was validated successfully
+          content:
+            text/plain:
+              schema:
+                type: string
+                example: sid:test-app.opdex.com/v1/auth?uid=nd0ZT7yrMpLPa207v1uGk0uu9NNcKILgn0HLJXpLQxGKuI9RKnugLdZPz_05a2tX-rCuN0tHM6z3_Qs1yi0Lyg&exp=1643082000
+        429:
+          $ref: "#/components/responses/TooManyRequests"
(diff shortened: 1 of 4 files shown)


mrtpain:
In postman you can use `Opdex > Platform API > Auth > Get Stratis Id` followed by `Opdex Wallet Helper > Login Direct` for any environment as long as you point to a valid FN with your wallet loaded to sign the message. This will set the access token in env variables

docs/openapi.yml
```diff
@@ -2767,13 +2767,33 @@ components:
...
+      tags:
+        - Authentication
+      summary: Get Stratis Id
+      description: >
+        Returns a new Stratis Id to sign and authenticate.
+        See the [specification](https://github.com/Opdex/SSAS) for further detail.
+      operationId: getStratisId
+      responses:
+        200:
+          description: Signature was validated successfully
```
drmathias, line 2780:
```suggestion
          description: Stratis Id was created
```

src/Opdex.Platform.WebApi/Controllers/AuthController.cs
```diff
@@ -25,35 +24,90 @@ namespace Opdex.Platform.WebApi.Controllers;
...
+                                                                 CancellationToken cancellationToken)
+    {
+        var created = Uri.TryCreate(_baseUri, UriKind.Absolute, out var callbackUri);
+
+        if (!created) throw new Exception("Unable to create callback URI");
+
+        (string _, string bearerToken) = await ValidateStratisSignature(callbackUri, query, body, cancellationToken);
+
+        return Ok(bearerToken);
+    }
```
drmathias, line 82: Nice, I like this.

drmathias (approved):

PR #193, opened 2022-01-25
[PAPI-344] Blocks endpoints

drmathias:
Adds endpoints to retrieve indexed blocks

* `GET /blocks`
* `GET /blocks/{height}`


src/Opdex.Platform.WebApi/Controllers/BlocksController.cs
```diff
@@ -0,0 +1,45 @@
...
+    private readonly IMapper _mapper;
+    private readonly IMediator _mediator;
+
+    public BlocksController(IMapper mapper, IMediator mediator)
+    {
+        _mapper = mapper;
+        _mediator = mediator;
+    }
+
+    [HttpGet]
```
mrtpain, line 27: nit - xml comments still here would be nice

mrtpain (approved):
Nice

PR #194, opened 2022-01-26
[PAPI-336] Inject Swagger Configi

mrtpain:
Inject swagger config within a JS file allowing for dynamic server urls when running the project vs static server urls within `openapi.yml`.

Now swagger environments will load correctly and openapi.yml imports default to testnet


mrtpain:
cc @drmathias relative URLs didn't work within ReadMe docs, it defaulted `https://example.com` there and calls would work. Found a happy medium, swagger ui config is included within an injected JS file where we can make the changes we need at runtime. 

Going to merge it but still tagged for the FYI

PR #197, opened 2022-01-27
[PAPI-353] Return decimals with precision of 8

drmathias:


mrtpain:
This looks like it will return out decimals as strings, I think we should keep usd decimal based values as numbers and try to find the root of the ones that are not returning with 8 decimals. 

I can help take a look around too, lmk if you know of any examples

drmathias:
> This looks like it will return out decimals as strings, I think we should keep usd decimal based values as numbers and try to find the root of the ones that are not returning with 8 decimals.
> 
> I can help take a look around too, lmk if you know of any examples

It returns out as a number, tried it out locally. It's why the `WriteRaw` method is used

mrtpain:
> It returns out as a number, tried it out locally. It's why the WriteRawValue method is used instead of WriteValue

I see, and I also see places where the root issue could be within snapshot models, so I can follow up separately there to make sure we're always rounding to 8 👍

mrtpain (approved):

PR #198, opened 2022-01-27
Exclude exceptions from sampling

drmathias:
Excludes exception logs from sampling. Followed instructions here: https://docs.microsoft.com/en-gb/azure/azure-monitor/app/sampling#adaptive-sampling


drmathias:
Merging and deploying to DEVNET to re-test max reorg alerts

PR #199, opened 2022-01-28
[PAPI-221] Cache-Control headers

drmathias:
- [x] Add Cache-Control header with max-age based on minimum expected new block time
- [x] Public/private directives depending on whether response data is user-specific
- [x] Response caching middleware
- [x] ~Respect Cache-Control related request headers~
- [x] OAS documentation

### DIFF
--- a/src/Opdex.Platform.Infrastructure/Data/DbContext.cs
+++ b/src/Opdex.Platform.Infrastructure/Data/DbContext.cs
@@ -5,6 +5,8 @@ using Dapper;
 using Microsoft.Extensions.Logging;
 using MySqlConnector;
 using Opdex.Platform.Infrastructure.Abstractions.Data;
+using Polly;
+using Polly.Retry;
 using System.Diagnostics;
 
 namespace Opdex.Platform.Infrastructure.Data;
@@ -13,6 +15,11 @@ namespace Opdex.Platform.Infrastructure.Data;
 // Ref: https://stackoverflow.com/questions/538060/proper-use-of-the-idisposable-interface
 public class DbContext : IDbContext
 {
+    private readonly AsyncRetryPolicy _retryPolicy = Policy
+        .Handle<MySqlException>(ex => ex.IsTransient)
+        .Or<TimeoutException>()
+        .WaitAndRetryAsync(3, x => TimeSpan.FromSeconds(x));
+
     private readonly ILogger<DbContext> _logger;
     private readonly IDatabaseSettings<MySqlConnection> _databaseSettings;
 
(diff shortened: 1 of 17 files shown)


drmathias:
It's probably not necessary to respect `Cache-Control` request headers as clients would not notice the difference and would only potentially take up additional server resources.

src/Opdex.Platform.WebApi/Caching/BlockBasedCacheControlFilter.cs
```diff
@@ -0,0 +1,54 @@
...
+    public async Task OnActionExecutionAsync(ActionExecutingContext context, ActionExecutionDelegate next)
+    {
+        var headers = context.HttpContext.Response.Headers;
+
+        // clear headers
+        headers.Remove(HeaderNames.CacheControl);
+        headers.Remove(HeaderNames.Pragma);
+
+        var status = await _mediator.Send(new GetIndexerStatusQuery(), CancellationToken.None);
+        if (status.Locked && status.Reason == IndexLockReason.Rewinding)
```
mrtpain, line 36: we may not want to cache when locked in general, think of if we pause indexing for a release. When its enabled, we might cache old results maybe?

src/Opdex.Platform.WebApi/Controllers/WalletsController.cs
```diff
@@ -93,6 +97,7 @@ public WalletsController(IApplicationContext context, IMapper mapper, IMediator
     /// <returns>Address balance summary.</returns>
     [HttpPost("{address}/balance/{token}")]
     [Authorize]
+    [CacheUntilNextBlock(CacheType.Public)]
```
mrtpain, line 100: Should this be private since there's an `Authorize` attribute here?

src/Opdex.Platform.Infrastructure/Data/DbContext.cs
```diff
@@ -13,6 +15,11 @@ namespace Opdex.Platform.Infrastructure.Data;
 // Ref: https://stackoverflow.com/questions/538060/proper-use-of-the-idisposable-interface
 public class DbContext : IDbContext
 {
+    private readonly AsyncRetryPolicy _retryPolicy = Policy
+        .Handle<MySqlException>(ex => ex.IsTransient)
+        .Or<TimeoutException>()
+        .WaitAndRetryAsync(3, x => TimeSpan.FromSeconds(x));
+
```
drmathias, line 22: Wooops, my bad I pushed the Dapper changes to this branch!
Anyway, it's a small change. The developers of the MySql extension were nice enough to provide a property that indicates if an exception may be transient, so it was a lot easier than I thought it would be.

src/Opdex.Platform.WebApi/Controllers/WalletsController.cs
drmathias: Private indicates the response might change per user. Any authorized user that calls this endpoint will get the same result. Potentially we could even remove the authorize attribute as it's not using any data from the token.

src/Opdex.Platform.WebApi/Caching/BlockBasedCacheControlFilter.cs
drmathias: Good point, removed the reason check

src/Opdex.Platform.WebApi/Controllers/WalletsController.cs
mrtpain: Makes sense, I kept it authorized only so users can't spam this endpoint with addresses from others, mainly so we're not reaching out to Cirrus if someone wants to hit this any # of times for other addresses within rate limits.

mrtpain (approved):

src/Opdex.Platform.WebApi/Controllers/WalletsController.cs
drmathias: fwiw this should no longer be much of a problem as the API server _should_ itself cache the response for 16 seconds

PR #201, opened 2022-02-01
[PAPI-356] Wrapped token queries

drmathias:
* Commands and queries to select and persist wrapped token chain mappings
* Returns out native token details in token responses
* Adds `Interflux` attribute
* Ability to filter by tokens native chain

### DIFF
--- a/src/Opdex.Platform.Application.Abstractions/Models/Tokens/MarketTokenDto.cs
+++ b/src/Opdex.Platform.Application.Abstractions/Models/Tokens/MarketTokenDto.cs
@@ -2,7 +2,8 @@ using Opdex.Platform.Common.Models;
 
 namespace Opdex.Platform.Application.Abstractions.Models.Tokens;
 
-public class MarketTokenDto : TokenDto
+public class
+    MarketTokenDto : TokenDto
 {
     public Address Market { get; set; }
     public Address LiquidityPool { get; set; }
--- a/src/Opdex.Platform.Application.Abstractions/Models/Tokens/MarketTokenDto.cs
+++ b/src/Opdex.Platform.Application.Abstractions/Models/Tokens/MarketTokenDto.cs
@@ -2,8 +2,7 @@ using Opdex.Platform.Common.Models;
 
 namespace Opdex.Platform.Application.Abstractions.Models.Tokens;
 
-public class
-    MarketTokenDto : TokenDto
+public class MarketTokenDto : TokenDto
 {
     public Address Market { get; set; }
     public Address LiquidityPool { get; set; }
--- a/src/Opdex.Platform.Application.Abstractions/Queries/Tokens/Wrapped/RetrieveTokenChainByTokenIdQuery.cs
+++ b/src/Opdex.Platform.Application.Abstractions/Queries/Tokens/Wrapped/RetrieveTokenChainByTokenIdQuery.cs
@@ -1,18 +1,16 @@
-using MediatR;
+using Opdex.Platform.Common.Queries;
 using Opdex.Platform.Domain.Models.Tokens;
 using System;
 
 namespace Opdex.Platform.Application.Abstractions.Queries.Tokens.Wrapped;
 
-public class RetrieveTokenChainByTokenIdQuery : IRequest<TokenChain>
+public class RetrieveTokenChainByTokenIdQuery : FindQuery<TokenChain>
 {
-    public RetrieveTokenChainByTokenIdQuery(ulong tokenId, bool findOrThrow = true)
+    public RetrieveTokenChainByTokenIdQuery(ulong tokenId, bool findOrThrow = true) : base(findOrThrow)
     {
         if (tokenId == 0) throw new ArgumentOutOfRangeException(nameof(tokenId), "Token id must be greater than 0.");
         TokenId = tokenId;
-        FindOrThrow = findOrThrow;
     }
 
     public ulong TokenId { get; }
-    public bool FindOrThrow { get; }
 }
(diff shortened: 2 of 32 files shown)


src/Opdex.Platform.Domain/Models/Tokens/TokenChain.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Opdex.Platform.Domain.Models.Tokens;
+
+public class TokenChain
+{
+    public TokenChain(ulong tokenId, ExternalChainType nativeChain, string nativeAddress)
+    {
+        if (tokenId == 0) throw new ArgumentOutOfRangeException(nameof(tokenId), "Token id must be greater than 0.");
+        if (!nativeChain.IsValid()) throw new ArgumentOutOfRangeException(nameof(nativeChain), "Chain type must be valid.");
+        if (!nativeAddress.HasValue()) throw new ArgumentNullException(nameof(nativeAddress), "Native address must be set.");
```
mrtpain, line 12: In the case that Bitcoin or a blockcore chain is ever supported w/ out smart contracts, native address should be allowed to be null.

src/Opdex.Platform.Application.Abstractions/Models/Tokens/MarketTokenDto.cs
```diff
@@ -2,8 +2,9 @@
 
 namespace Opdex.Platform.Application.Abstractions.Models.Tokens;
 
-public class MarketTokenDto : TokenDto
+public class
+    MarketTokenDto : TokenDto
```
mrtpain, line 6: nit -formatting

src/Opdex.Platform.Application.Abstractions/Queries/Tokens/Wrapped/RetrieveTokenChainByTokenIdQuery.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Opdex.Platform.Application.Abstractions.Queries.Tokens.Wrapped;
+
+public class RetrieveTokenChainByTokenIdQuery : IRequest<TokenChain>
+{
+    public RetrieveTokenChainByTokenIdQuery(ulong tokenId, bool findOrThrow = true)
+    {
+        if (tokenId == 0) throw new ArgumentOutOfRangeException(nameof(tokenId), "Token id must be greater than 0.");
+        TokenId = tokenId;
+        FindOrThrow = findOrThrow;
+    }
```
mrtpain, line 14: can use `: FindQuery<TokenChain>` base class instead

src/Opdex.Platform.Application/Assemblers/MarketTokenDtoAssembler.cs
```diff
@@ -34,9 +35,18 @@ public async Task<MarketTokenDto> Assemble(MarketToken token)
...
         marketTokenDto.Attributes = tokenAttributes.Select(attr => attr.AttributeType);
 
+        var tokenChain = await _mediator.Send(new RetrieveTokenChainByTokenIdQuery(token.Id, false), CancellationToken.None);
+        if (tokenChain is not null)
+        {
+            marketTokenDto.NativeToken = new WrappedTokenDetailsDto
+            {
+                Chain = tokenChain.NativeChain,
+                Address = tokenChain.NativeAddress
+            };
```
mrtpain, line 47: Might be worth adding to the application mapper profile so this and `TokenDtoAssembler` can share

src/Opdex.Platform.Application/Handlers/Tokens/Wrapped/MakeTokenChainCommandHandler.cs
```diff
@@ -0,0 +1,23 @@
...
+    private readonly IMediator _mediator;
+
+    public MakeTokenChainCommandHandler(IMediator mediator)
+    {
+        _mediator = mediator ?? throw new ArgumentNullException(nameof(mediator));
+    }
+
+    public async Task<ulong> Handle(MakeTokenChainCommand request, CancellationToken cancellationToken)
+    {
+        return await _mediator.Send(new PersistTokenChainCommand(request.Chain), cancellationToken);
```
mrtpain, line 21: Something we should probably look into for other make commands but I think we should use `CancellationToken.None` here just incase.

src/Opdex.Platform.Infrastructure.Abstractions/Data/Queries/Tokens/Wrapped/SelectTokenChainByTokenIdQuery.cs
```diff
@@ -0,0 +1,18 @@
+using MediatR;
+using Opdex.Platform.Domain.Models.Tokens;
+using System;
+
+namespace Opdex.Platform.Infrastructure.Abstractions.Data.Queries.Tokens.Wrapped;
+
+public class SelectTokenChainByTokenIdQuery : IRequest<TokenChain>
```
mrtpain, line 7: can use `FindQuery<TokenChain>`

src/Opdex.Platform.Infrastructure.Abstractions/Data/Queries/Tokens/TokensCursor.cs
```diff
@@ -158,6 +170,20 @@ public enum TokenAttributeFilter
...
+}
+
+/// <summary>
+/// Selectable options for the token native chain
+/// </summary>
+public enum ChainType
+{
+    Cirrus = 0,
+    Ethereum = 1
 }
```
mrtpain, line 187: Something we should come back to for all paginated endpoints but we can probably start here with this one.

Currently, these external client filter values match up with our database and internal enum values but this is an easy thing to miss and assumes the developer will use consistent values. An example of different external vs internal values is filtering token attributes, security attributes are not filterable, so if developers ordered enum values sequentially without knowing, it would create a mismatch and query for something else. 

We should create mappers for these public filter types to out internal database types so this won't be a future problem.

src/Opdex.Platform.Domain/Models/Tokens/TokenChain.cs
drmathias: Good spot. This is an oversight, my intention was that for the base token of the chain, the `NativeAddress` would be empty/null

PR #204, opened 2022-02-03
[PAPI-358] Persist interflux token mappings

drmathias:
- [x] Persist on add token endpoint
- [x] Persist on liquidity pool indexing
- [x] Finish testing


src/Opdex.Platform.Common/Models/Address.cs
```diff
@@ -1,8 +1,11 @@
+using Opdex.Platform.Common.Converters;
 using Opdex.Platform.Common.Extensions;
 using System;
+using System.ComponentModel;
 
 namespace Opdex.Platform.Common.Models;
 
+[TypeConverter(typeof(AddressTypeConverter))]
```
drmathias, line 8: This is added to allow `Configuration` values to bind to an `Address` type

src/Opdex.Platform.WebApi/appsettings.json
```diff
@@ -48,6 +48,9 @@
     "ApiKey": "",
     "HeaderName": "X-CMC_PRO_API_KEY"
   },
+  "InterfluxConfiguration": {
+    "MultiSigContractAddress": "PBHvTPaLKo5cVYBFdTfTgtjqfybLMJJ8W5"
```
drmathias, line 52: Will remove

src/Opdex.Platform.Application/EntryHandlers/Tokens/CreateAddTokenCommandHandler.cs
```diff
@@ -19,42 +24,79 @@ public class CreateAddTokenCommandHandler : IRequestHandler<CreateAddTokenComman
...
         }
         catch (Exception)
         {
             throw new InvalidDataException("token", "Unable to validate SRC token.");
         }
 
+        var interfluxSummary = await _mediator.Send(new CallCirrusGetInterfluxTokenContractSummaryQuery(request.Token, latestBlock.Height), CancellationToken.None);
+        if (interfluxSummary is not null && !ValidateWrappedToken(interfluxSummary))
+        {
+            throw new InvalidDataException("token", "Wrapped token must be owned by Interflux multisig wallet.");
```
drmathias, line 65: This feels the best thing to do, though feeling conflicted about it. Any token which is created via `CreateTokenCommandHandler` will still be persisted, just without the `TokenChain` interflux mapping.

src/Opdex.Platform.Application/EntryHandlers/Tokens/CreateAddTokenCommandHandler.cs
```diff
@@ -53,8 +61,33 @@ public async Task<TokenDto> Handle(CreateAddTokenCommand request, CancellationTo
                           summary.TotalSupply.Value, latestBlock.Height);
 
         var tokenId = await _mediator.Send(new MakeTokenCommand(token, latestBlock.Height));
-        if (tokenId == 0) throw new Exception("Something went wrong when indexing the token.");
+        if (tokenId == 0) _logger.LogError("Something went wrong indexing the token");
```
mrtpain, line 64: I think throwing here is necessary still or we'll try to insert a wrapped token with tokenId 0

src/Opdex.Platform.Infrastructure/Clients/CirrusFullNodeApi/Handlers/Tokens/CallCirrusGetInterfluxTokenContractSummaryQueryHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public async Task<InterfluxTokenContractSummary> Handle(CallCirrusGetInterfluxTokenContractSummaryQuery request, CancellationToken cancellationToken)
+    {
+        try
+        {
+            var ownerRequest = new LocalCallRequestDto(request.Token, request.Token,
+                $"get_{InterfluxTokenConstants.Properties.Owner}", request.BlockHeight);
+            var ownerResponse = await _smartContractsModule.LocalCallAsync(ownerRequest, cancellationToken);
+            var owner = ownerResponse.DeserializeValue<Address>();
```
mrtpain, line 31: Since this is making multiple calls and we're checking this for all created tokens, we should fail fast rather than throw exceptions within `summary.SetInterfluxDetails` to catch then return null.

If the Owner is null or Address.Zero - return null immediately
if the NativeChain isn't valid - return null immediately

src/Opdex.Platform.WebApi/Mappers/TrustedBridgeMappingAction.cs
```diff
@@ -0,0 +1,22 @@
+using AutoMapper;
+using Opdex.Platform.Application.Abstractions.Models.Tokens;
+using Opdex.Platform.Infrastructure.Abstractions.Clients.CirrusFullNodeApi;
+using Opdex.Platform.WebApi.Models.Responses.Tokens;
+using System;
+
+namespace Opdex.Platform.WebApi.Mappers;
+
+public class TrustedBridgeMappingAction : IMappingAction<WrappedTokenDetailsDto, WrappedTokenDetailsResponseModel>
```
mrtpain, line 9: Fancy, this is cool

src/Opdex.Platform.WebApi/Mappers/PlatformWebApiMapperProfile.cs
```diff
@@ -59,8 +59,11 @@ public class PlatformWebApiMapperProfile : Profile
     public PlatformWebApiMapperProfile()
     {
         CreateMap<WrappedTokenDetailsDto, WrappedTokenDetailsResponseModel>()
+            .ForMember(dest => dest.Custodian, opt => opt.MapFrom(src => src.Custodian))
             .ForMember(dest => dest.Chain, opt => opt.MapFrom(src => src.Chain))
             .ForMember(dest => dest.Address, opt => opt.MapFrom(src => src.Address))
+            .ForMember(dest => dest.ModifiedBlock, opt => opt.MapFrom(src => src.ModifiedBlock))
```
mrtpain, line 65: I think we might do this somewhere else too, I don't recall exactly, but it might make sense to just return CreatedBlock too everywhere so its consistent for clients, even if it won't change or matches its parent. 

Not a huge deal but maybe something to consider

src/Opdex.Platform.Infrastructure.Abstractions/Clients/CirrusFullNodeApi/InterfluxConfiguration.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Opdex.Platform.Infrastructure.Abstractions.Clients.CirrusFullNodeApi;
+
+public class InterfluxConfiguration : IValidatable
+{
+    public Address MultiSigContractAddress { get; set; }
+
+    public void Validate()
+    {
+        if (MultiSigContractAddress == Address.Empty || MultiSigContractAddress == Address.Cirrus)
```
mrtpain, line 13: Could also check `.IsZero`

src/Opdex.Platform.Domain/Models/Tokens/InterfluxTokenContractSummary.cs
```diff
@@ -0,0 +1,27 @@
...
+using Opdex.Platform.Common.Models;
+using System;
+
+namespace Opdex.Platform.Domain.Models.Tokens;
+
+public class InterfluxTokenContractSummary
+{
+    public InterfluxTokenContractSummary(ulong blockHeight)
+    {
+    }
```
mrtpain, line 11: ✂️

src/Opdex.Platform.Domain/Models/Tokens/TokenWrapped.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    public ulong Id { get; }
+    public ulong TokenId { get; }
+    public Address Owner { get; private set; }
+    public ExternalChainType NativeChain { get; }
+    public string NativeAddress { get; }
+
+    public void SetOwner(Address owner, ulong blockHeight)
+    {
+        if (owner == Address.Empty) throw new ArgumentNullException(nameof(owner), "Owner address must be set.");
```
mrtpain, line 40: Exception message can change to "Owner address already set."

src/Opdex.Platform.Application/EntryHandlers/Tokens/CreateTokenCommandHandler.cs
```diff
@@ -14,40 +17,57 @@ namespace Opdex.Platform.Application.EntryHandlers.Tokens;
...
+
+        var interfluxSummary = await _mediator.Send(new CallCirrusGetInterfluxTokenContractSummaryQuery(request.Token, request.BlockHeight), CancellationToken.None);
+        if (interfluxSummary is not null)
+        {
+            attributes = request.Attributes.Append(TokenAttributeType.Interflux);
+
+            var tokenWrapped = new TokenWrapped(tokenId, interfluxSummary.Owner, interfluxSummary.NativeChain, interfluxSummary.NativeAddress, request.BlockHeight);
+            var tokenWrappedId = await _mediator.Send(new MakeTokenWrappedCommand(tokenWrapped), CancellationToken.None);
+            if (tokenWrappedId == 0) _logger.LogError("Something went wrong indexing the wrapped token mapping");
+        }
```
mrtpain, line 63: I think it might be worth making a CreateTokenWrappedCommand here and a `CreateTokenAttributesCommand` for below to share between the create token command handlers.  Not exactly pressing for it, but it would be nicer to maintain.

src/Opdex.Platform.Application/EntryHandlers/Tokens/CreateAddTokenCommandHandler.cs
drmathias: True, I've added the exception back.

src/Opdex.Platform.Infrastructure/Clients/CirrusFullNodeApi/Handlers/Tokens/CallCirrusGetInterfluxTokenContractSummaryQueryHandler.cs
drmathias: Resolved this and while at it, added a `TryDeserializeValue` instead so we're not throwing exceptions. Although, I'm not returning null if the Owner is `Address.Zero` since there's a method on the contract to revoke ownership, which sets it to `Address.Zero`. In which case we'd still want to index this

src/Opdex.Platform.Application/EntryHandlers/Tokens/CreateTokenCommandHandler.cs
drmathias: I'm not sure this would make maintenance any easier tbh. Domain events might be nice to implement and would ease maintenance a lot, but that's a big time investment.

src/Opdex.Platform.Infrastructure/Clients/CirrusFullNodeApi/Handlers/Tokens/CallCirrusGetInterfluxTokenContractSummaryQueryHandler.cs
```diff
@@ -23,31 +24,24 @@ public CallCirrusGetInterfluxTokenContractSummaryQueryHandler(ISmartContractsMod
...
-            var nativeAddressRequest = new LocalCallRequestDto(request.Token, request.Token,
-                $"get_{InterfluxTokenConstants.Properties.NativeAddress}", request.BlockHeight);
-            var nativeAddressResponse =
-                await _smartContractsModule.LocalCallAsync(nativeAddressRequest, cancellationToken);
-            var nativeAddress = nativeAddressResponse.DeserializeValue<string>();
+        var nativeAddressRequest = new LocalCallRequestDto(request.Token, request.Token,
+            $"get_{InterfluxTokenConstants.Properties.NativeAddress}", request.BlockHeight);
+        var nativeAddressResponse =
+            await _smartContractsModule.LocalCallAsync(nativeAddressRequest, cancellationToken);
+        if (!nativeAddressResponse.TryDeserializeValue<string>(out var nativeAddress)) return null;
```
mrtpain, line 41: Null is ok here

drmathias: Yes, this will only evaluate to false if the value cannot be deserialized. When `TryDeserializeValue` is called trying to deserialise to a reference type, `null` is a valid value and it'll return true.

PR #205, opened 2022-02-04
[PAPI-364] Fix token paging backwards results

drmathias:
Resolves bug where a query to retrieve a previous page of token results with an order by filter would fail

### DIFF
--- a/src/Opdex.Platform.Infrastructure/Data/Handlers/Tokens/SelectTokensWithFilterQueryHandler.cs
+++ b/src/Opdex.Platform.Infrastructure/Data/Handlers/Tokens/SelectTokensWithFilterQueryHandler.cs
@@ -198,7 +198,7 @@ public class SelectTokensWithFilterQueryHandler : IRequestHandler<SelectTokensWi
             direction = Enum.GetName(typeof(SortDirectionType), request.Cursor.SortDirection);
 
         // Order the rows by the preferred, indexed column or tokenId by default
-        var orderBy = OrderByBuilder(request.Cursor.OrderBy, direction, reverse: false);
+        var orderBy = InnerOrderByBuilder(request.Cursor.OrderBy, direction);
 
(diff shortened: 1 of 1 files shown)


src/Opdex.Platform.Infrastructure/Data/Handlers/Tokens/SelectTokensWithFilterQueryHandler.cs
```diff
@@ -215,24 +215,35 @@ private static string QueryBuilder(SelectTokensWithFilterQuery request)
...
-        var tokenPrefix = reverse ? "r" : "t";
+        var tokenIdWithDirection = $"t.{nameof(TokenEntity.Id)} {direction}";
 
-        var tokenIdWithDirection = $"{tokenPrefix}.{nameof(TokenEntity.Id)} {direction}";
+        return cursorOrderBy switch
+        {
+            TokenOrderByType.Name => $" ORDER BY t.{nameof(TokenEntity.Name)} {direction}, {tokenIdWithDirection}",
+            TokenOrderByType.Symbol => $" ORDER BY t.{nameof(TokenEntity.Symbol)} {direction}, {tokenIdWithDirection}",
+            TokenOrderByType.PriceUsd => $" ORDER BY AVG(ts.{nameof(TokenSummaryEntity.PriceUsd)}) {direction}, {tokenIdWithDirection}",
+            TokenOrderByType.DailyPriceChangePercent => $" ORDER BY AVG(ts.{nameof(TokenSummaryEntity.DailyPriceChangePercent)}) {direction}, {tokenIdWithDirection}",
```
mrtpain, line 232: Should this `AVG` be here?

drmathias: Yes the inner query has to order by the AVG; the outer query orders the already-averaged column returned from the inner query

mrtpain (approved):

PR #206, opened 2022-02-06
[PAPI-357] Track Interflux token ownership and supply

drmathias:
* Tracking of `OwnershipTransferred` and `SupplyChangeLog`
* Mapping and returning logs as events


src/Opdex.Platform.Infrastructure/PlatformInfrastructureMapperProfile.cs
```diff
@@ -330,6 +334,10 @@ public PlatformInfrastructureMapperProfile()
                     nameof(VaultProposalVoteLog) => new VaultProposalVoteLog(src.Log.Data, src.Address, src.SortOrder),
                     nameof(VaultProposalWithdrawVoteLog) => new VaultProposalWithdrawVoteLog(src.Log.Data, src.Address, src.SortOrder),
 
+                    // Interflux Tokens
+                    "OwnershipTransferred" => new OwnershipTransferredLog(src.Log.Data, src.Address, src.SortOrder),
```
mrtpain, line 338: 😭

src/Opdex.Platform.Common/Enums/TransactionEventType.cs
```diff
@@ -47,5 +47,9 @@ public enum TransactionEventType
     VaultProposalPledgeEvent = 28,
     VaultProposalWithdrawPledgeEvent = 29,
     VaultProposalVoteEvent = 30,
-    VaultProposalWithdrawVoteEvent = 31
+    VaultProposalWithdrawVoteEvent = 31,
+
+    // Interflux tokens
+    OwnershipTransferredEvent = 32,
+    SupplyChangeEvent = 33,
```
mrtpain, line 54: Side note, maybe you could add to this PR. 

These enums are off by 1 from the `TransactionLogType` because a `ReservesEvent` doesn't exist in the Liquidity Pools section. Would need a response model and probably a DTO assembler so can be done as a separate task too, up to you.

mrtpain (approved):
lgtm, one comment about the ReservesEvent but can be a separate task if you prefer. 

Also, are openapi.yml changes needed for the new events?

mrtpain:
Just thought about and noticed `TransactionEventExtensions.cs` needs to be updated with the new logs/event mapping for search filters. 

That extension method can probably return a single `TransactionLogType` now too rather than an IEnumerable. It used to map one to many but not anymore.

src/Opdex.Platform.Common/Enums/TransactionEventType.cs
drmathias: Happy to do this - I think we'll need to rewind testnet back to the first block if we want the indexer to pick up these events

drmathias: Appears to have resolved the issue with missing `sortOrder` values

src/Opdex.Platform.Infrastructure/PlatformInfrastructureMapperProfile.cs
```diff
@@ -750,5 +750,56 @@ public PlatformInfrastructureMapperProfile()
...
+                    TransactionEventType.SetPendingMarketOwnershipEvent => TransactionLogType.SetPendingMarketOwnershipLog,
+                    TransactionEventType.ChangeMarketPermissionEvent => TransactionLogType.ChangeMarketPermissionLog,
+                    TransactionEventType.CreateLiquidityPoolEvent => TransactionLogType.CreateLiquidityPoolLog,
+                    // Liquidity Pools
+                    TransactionEventType.SwapEvent => TransactionLogType.SwapLog,
+                    TransactionEventType.AddLiquidityEvent => TransactionLogType.MintLog,
+                    TransactionEventType.RemoveLiquidityEvent => TransactionLogType.BurnLog,
+                    TransactionEventType.StartStakingEvent => TransactionLogType.StartStakingLog,
+                    TransactionEventType.StopStakingEvent => TransactionLogType.StopStakingLog,
+                    TransactionEventType.CollectStakingRewardsEvent => TransactionLogType.CollectStakingRewardsLog,
```
mrtpain, line 775: We can add the ReservesEvent here so it can be searched for

mrtpain:
lgtm but were you thinking of adding the Reserves Event in another PR or including within this one?

drmathias:
> lgtm but were you thinking of adding the Reserves Event in another PR or including within this one?

The reserves event is included in this PR, named as `ReservesChangeEvent`

PR #207, opened 2022-02-06
[PAPI-338] Suppress manual rewind tx notifications

drmathias:


src/Opdex.Platform.WebApi/Controllers/IndexerController.cs
```diff
@@ -92,6 +92,8 @@ public async Task<IActionResult> ResyncFromDeployment(ResyncFromDeploymentReques
     [Authorize(Policy = "AdminOnly")]
     public async Task<ActionResult> Rewind(RewindRequest request, CancellationToken cancellationToken)
     {
+        var startingBlock = await _mediator.Send(new RetrieveLatestBlockQuery(), CancellationToken.None);
```
mrtpain, line 95: Nice

mrtpain (approved):
Nice work, I like that we the `notifyAfterBlock` working from manual rewinds but notifying all during automated rewinds. 👍

PR #208, opened 2022-02-06
[PAPI-365] CoinMarketCap price feed feature flag

drmathias:
* Defaults to using CoinGecko price feed
* Feature flag to use CoinMarketCap price feed instead

Realise toggling between feeds is not the desired behaviour in terms of resiliency. For now this is a step toward allowing the use of CoinGecko and removes the need to use paid dependencies to run the API.

The idea is that we can use `FiatPriceFeed` to write our logic that will aggregate USD pricing data.


mrtpain (approved):
NIce lgtm

PR #209, opened 2022-02-07
[PAPI-354] Market and liquidity pool response cleanup

drmathias:
Cleans up the market and liquidity pool responses so that they are structured in the same way. Makes it easier to consume data and store only what is cared for.

Also includes a fix so that no staking summary is returned for standard markets.

### DIFF
--- a/docs/openapi.yml
+++ b/docs/openapi.yml
@@ -1890,6 +1890,8 @@ components:
           $ref: "#/components/schemas/tokenResponse"
         lp:
           $ref: "#/components/schemas/tokenResponse"
+        staking:
+          $ref: "#/components/schemas/tokenResponse"
     liquidityPoolSummary:
       type: object
       properties:
@@ -1944,8 +1946,6 @@ components:
     liquidityPoolStakingSummary:
       type: object
       properties:
-        token:
-          $ref: "#/components/schemas/marketTokenResponse"
         weight:
           $ref: "#/components/schemas/fixedDecimal"
           description: Total number of tokens staking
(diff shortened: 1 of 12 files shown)


drmathias:
@mrtpain Reminder, this requires changes in the UI project

docs/openapi.yml
```diff
@@ -2101,6 +2098,16 @@ components:
...
           description: Financial summary for the market
+    marketTokenBreakdown:
+      type: object
+      description: Details of tokens that are tied to a market
+      properties:
+        crs:
+          $ref: "#/components/schemas/tokenResponse"
+          description: Details of the CRS token
+        staking:
+          $ref: "#/components/schemas/tokenResponse"
```
mrtpain, line 2109: this one would be a marketTokenResponse

docs/openapi.yml
```diff
@@ -3451,6 +3458,22 @@ paths:
...
+                      staking:
+                        address: tTTuKbCR2UnsEByXBp1ynBz91J2yz63h1c
+                        name: Testnet Opdex Token
+                        symbol: TODX
+                        decimals: 8
+                        sats: 100000000
+                        totalSupply: "400000000.00000000"
+                        attributes:
+                          - NonProvisional
+                          - Staking
```
mrtpain, line 3470: market token response would have market and liquidity pool properties

mrtpain (approved):
Couple comments, will make a PR for the client

docs/openapi.yml
drmathias: I'm using `MarketTokenDto` so that we get the market token summary, then mapping to `TokenResponseModel` so that the `market` and `liquidityPool` address are not repeated in the response

docs/openapi.yml
drmathias: See above comment, the market response already includes the market address so using `tokenResponse` here avoids that

mrtpain: Ok, my thinking was using the MarketToken response for the market's staking token because it also includes the liquidity pool address even though we don't need the market address that comes with it. 

Our client won't look up a staking pool directly like that but others might

PR #210, opened 2022-02-08
Standardize created and modified block in responses

mrtpain:
PR that pairs with https://github.com/Opdex/opdex-v1-ui/pull/90 for decoupling UI models from API response models. 

---

This PR modifies some models that return `ModifiedBlock` but not `CreatedBlock`. Now returning CreatedBlock wherever ModifiedBlock is returned to standardize these responses to ease client integrations allowing inheritance when not using any sort of SDK.

**Todo**
- [ ] OpenApi Examples


drmathias (approved):
LGTM as long as the OAS examples are updated

PR #219, opened 2022-02-22
[PAPI-366] Selectively process smart contract txs

drmathias:
Parses the block receipt to selectively process smart contract call transactions only


drmathias:
Will do the retries in a separate PR

src/Opdex.Platform.Infrastructure.Abstractions/Clients/CirrusFullNodeApi/Models/RawTransactionDto.cs
```diff
@@ -10,14 +13,61 @@ public RawTransactionDto()
...
+            var isValidScript = byte.TryParse(Hex[..2], NumberStyles.HexNumber, CultureInfo.InvariantCulture, out var firstByte);
+            if (!isValidScript) return default;
+            var smartContractScriptType = ((ScOpcodeType)firstByte);
+            var isValidSmartContractScript = smartContractScriptType.IsValid();
+            return isValidSmartContractScript ? smartContractScriptType : default;
+        }
+    }
+}
+
 public class ScriptPubKeyDto
```
mrtpain, line 71: Can this class inherit `ScriptSignDto` here to save some duplicate code, or another abstract base class altogether for both `ScriptSigDto` and `ScriptPubKeyDto`?

src/Opdex.Platform.Infrastructure.Abstractions/Clients/CirrusFullNodeApi/Models/RawTransactionDto.cs
```diff
@@ -10,14 +13,61 @@ public RawTransactionDto()
...
+     *
+     * Coinbase and SegWit transactions do not contain a script sig in the input. Therefore, we filter these out.
+     * We filter out transactions where the script sig uses nonstandard smart contract OP codes (internal transfers).
+     * Finally, we return transactions where the script pub key uses nonstandard smart contract OP codes.
+     *
+     */
+    public bool IsSmartContractCall => Vout.Any(v => v.ScriptPubKey.SmartContractScriptType == ScOpcodeType.OP_CALLCONTRACT);
+    public bool IsSmartContractCreate => Vout.Any(v => v.ScriptPubKey.SmartContractScriptType == ScOpcodeType.OP_CREATECONTRACT);
+
+    public bool IsExternalSmartContractTransfer => Vin.All(v => v.ScriptSig is not null && !v.ScriptSig.IsSmartContractScript);
```
mrtpain, line 37: I know we've discussed this earlier but a small comment about this pointing to what you've shared with me could go a long way in the future if/when we need to come back to this code. 

Might be overkill but maybe not a bad idea to reference the previous links you've shared too:
https://github.com/stratisproject/StratisFullNode/blob/master/src/Stratis.SmartContracts.CLR/ContractExecutor.cs#L90
https://chainz.cryptoid.info/cirrus-test/block.dws?3434decc0e31d8e0a5cccb662b73ac225beaf259c7e8eceec69f83c71c69f7e2.htm

mrtpain (approved):
lgtm, nice work, hope it works and we don't miss txs 🤞 
We can leave this in dev for a few days once merged and do some testing of different transaction types to make sure things aren't missed prior to moving to testnet

src/Opdex.Platform.Infrastructure.Abstractions/Clients/CirrusFullNodeApi/Models/RawTransactionDto.cs
drmathias: Updated to inherit a generic `ScriptDto` amended the method to get the smart contract script type to be an extension method.

src/Opdex.Platform.Infrastructure.Abstractions/Clients/CirrusFullNodeApi/Models/RawTransactionDto.cs
drmathias: Added some additional comments as you suggested

drmathias:
> lgtm, nice work, hope it works and we don't miss txs 🤞 We can leave this in dev for a few days once merged and do some testing of different transaction types to make sure things aren't missed prior to moving to testnet

I tested it a bit locally, so I'm hopeful that it'll work smoothly :)

PR #222, opened 2022-02-28
[PAPI-387] Upgrade AutoMapper to v11

drmathias:
AutoMapper v11 has breaking changes. See [upgrade guide](https://docs.automapper.org/en/latest/11.0-Upgrade-Guide.html#forallothermembers-was-removed). It is currently the only dependency that we have with breaking changes.

We can no longer use `ForAllOtherMembers`, so instead I've opted to replace this with `ValidateMemberList(MemberList.None)`, to keep the behaviour where unspecified properties do not cause a runtime exception.

Going forward, I believe it will be best to aim to explicitly ignore unmapped properties. This is required for properties where we have two properties of the same name, but with different types (for example, where we have an assembler, the mapping fails for Amount -> FixedDecimal and must be ignored).

### DIFF
--- a/src/Opdex.Platform.Application/Assemblers/VaultCertificateDtoAssembler.cs
+++ b/src/Opdex.Platform.Application/Assemblers/VaultCertificateDtoAssembler.cs
@@ -1,10 +1,14 @@
 using AutoMapper;
 using MediatR;
 using Opdex.Platform.Application.Abstractions.Models.Vaults;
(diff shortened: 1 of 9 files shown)


drmathias:
No other dependencies have breaking changes and have been updated on the `feature/PAPI-386_...` branch.

mrtpain (approved):

PR #223, opened 2022-03-02
[PAPI-389] Use new node endpoint to determine InterFlux token trust

drmathias:


src/Opdex.Platform.Common/Models/Address.cs
```diff
@@ -56,7 +56,7 @@ public Address(string value)
 
     public override int GetHashCode()
     {
-        return Value.GetHashCode();
+        return Value is null ? "".GetHashCode() : Value.GetHashCode();
```
mrtpain, line 59: Doesn't make a difference but could probably use `string.Empty.GetHashCode();`

src/Opdex.Platform.Application/PlatformApplicationServiceCollectionExtensions.cs
```diff
@@ -684,4 +687,10 @@ private static IServiceCollection AddAssemblers(this IServiceCollection services
 
         return services;
     }
+
+    private static IServiceCollection AddCacheServices(this IServiceCollection services)
```
mrtpain, line 691: Curious about naming, "Cache" doesn't seem very fitting here right?

src/Opdex.Platform.Application/Cache/WrappedTokenTrustValidator.cs
```diff
@@ -0,0 +1,35 @@
...
+
+    public async ValueTask<bool> Validate(Address token, CancellationToken cancellationToken = default)
+    {
+        if (_trustedTokens is null)
+        {
+            var supportedTokens = await _supportedContractsModule.GetList(cancellationToken);
+            _trustedTokens = supportedTokens.Select(t => t.Src20Address).ToImmutableHashSet();
+        }
+
+        return _trustedTokens.Contains(token);
```
mrtpain, line 33: Don't think _trustedTokens would ever _not_ be null right?

src/Opdex.Platform.Common/Models/Address.cs
drmathias: Updated

src/Opdex.Platform.Application/PlatformApplicationServiceCollectionExtensions.cs
mrtpain: That's going to require a redeployment or restart of the API when we deploy a new FN version. 

Thinking now that the list is retrieved from the FN, we add a `Trusted` DB column to the `token_wrapped` table and set it once during token creation. Then we don't need to cache anything and the list would only be retrieved when adding a new token.

mrtpain (approved):
lgtm, nice work as always

drmathias:
Will hold off from merging until we're able to update the Cirrus node

mrtpain:
@drmathias I added this commit https://github.com/Opdex/opdex-v1-api/pull/223/commits/b7590f1fca6d8e2295a0eafc52facfdbc74bdef9 to adjust the Interflux supported token model to match Stratis FN, also added a temporary static list so we can create liquidity pools while waiting for Stratis.Bitcoin package updates.

PR #224, opened 2022-03-03
[PAPI-388] OpenAPI fixes

drmathias:
* Fixes Address RegEx
* Start/End datetime no longer marked required on history endpoints (not required when cursor is used)
* Added `CreatedBlock` option for market/token/lp order by param and marked as default. Ordering by id has the same effect so CreatedBlock is used, as it is not abstracted from the response


src/Opdex.Platform.Infrastructure.Abstractions/Data/Queries/LiquidityPools/LiquidityPoolsCursor.cs
```diff
@@ -261,7 +261,7 @@ public enum LiquidityPoolOrderByType
     /// <summary>
     /// The default status of any is ignored in filter.
     /// </summary>
-    Any = 0,
+    CreatedBlock = 0,
```
mrtpain, line 264: liquidity pools and transactions filter endpoints probably need a similar fix

mrtpain (commented):
I put a comment on this task in Jira, will put it here too. Can you remove the Vault Ownership events from OpenApi transaction event filter options. Since they no longer exist they throw an error.

src/Opdex.Platform.Infrastructure.Abstractions/Data/Queries/LiquidityPools/LiquidityPoolsCursor.cs
drmathias: Believe I've covered all of the enum options which don't have a default value

src/Opdex.Platform.Common/Enums/TransactionEventType.cs
```diff
@@ -16,8 +16,8 @@ public enum TransactionEventType
     // Liquidity pool logs
     AddLiquidityEvent = 8,
     RemoveLiquidityEvent = 9,
-    SwapEvent = 10,
-    ReservesChangeEvent = 11,
+    ReservesChangeEvent = 10,
+    SwapEvent = 11,
     StartStakingEvent = 12,
```
mrtpain, line 21: Not sure about this, will need to check out mappings. Was there a need for this change? 

Our TransactionLogTypes and database have Swap at 10 and Reserves at 11.

drmathias: My bad, I changed this to match the order they were listed in OpenAPI. Should have updated the order in OpenAPI, if anything. Reverted

mrtpain (approved):

PR #228, opened 2022-03-08
[PAPI-397] Persist Attributes

mrtpain:
When a pool was created, most of the time, we'd already know about the token through our validation process. 

When creating the pool and attempting to persist the SRC token, if it already exists, try to persist additional attributes rather than returning out early.

`MakeTokenAttributeCommand` will only attempt to persist attributes that are not already applied to the given token. 

Hide whitespace for easier review.

### DIFF
--- a/src/Opdex.Platform.Application/EntryHandlers/Tokens/CreateTokenCommandHandler.cs
+++ b/src/Opdex.Platform.Application/EntryHandlers/Tokens/CreateTokenCommandHandler.cs
@@ -8,6 +8,7 @@ using Opdex.Platform.Common.Enums;
 using Opdex.Platform.Domain.Models.Tokens;
 using Opdex.Platform.Infrastructure.Abstractions.Clients.CirrusFullNodeApi.Queries.Tokens;
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


drmathias:
All of this _should_ already get persisted when adding the token. The only reason I can think it wouldn't be is if tokens were added before the token_wrapped changes were made.

mrtpain:
> All of this _should_ already get persisted when adding the token. The only reason I can think it wouldn't be is if tokens were added before the token_wrapped changes were made.

`Provisional` and `NonProvisional` attributes are based on the tokens and their relations with liquidity pools. OLPT don't get verified or added before a pool is created so those were fine. But SRC tokens were missing the `NonProvisional` attribute since our DB already knew about the token, just didn't know its purpose to apply that attribute.

src/Opdex.Platform.Application/EntryHandlers/Tokens/CreateTokenCommandHandler.cs
```diff
@@ -37,20 +37,25 @@ public async Task<ulong> Handle(CreateTokenCommand request, CancellationToken ca
...
+                          summary.TotalSupply.GetValueOrDefault(),
+                          request.BlockHeight);
 
-        tokenId = await _mediator.Send(new MakeTokenCommand(token, request.BlockHeight));
-        if (tokenId == 0) _logger.LogError("Something went wrong indexing the token");
+        tokenId = await _mediator.Send(new MakeTokenCommand(token, request.BlockHeight), CancellationToken.None);
+
+        if (tokenId == 0)
+        {
+            _logger.LogError("Something went wrong indexing the token");
```
drmathias, line 56: I know that we weren't already, but we should log the token address as a custom property

drmathias (approved):
LGTM

src/Opdex.Platform.Application/EntryHandlers/Tokens/CreateTokenCommandHandler.cs
mrtpain: Added this, there are many other logs that could use some updates similar to this as well. I didn't do them but we can come back.

PR #230, opened 2022-03-09
[PAPI-390] Maintenance lock filter

drmathias:
* Applies maintenance lock middleware selectively to controllers, excluding: Auth, Indexer and Status
* Adds a field to the status endpoint to include the maintenance status


src/Opdex.Platform.WebApi/Controllers/StatusController.cs
```diff
@@ -27,7 +30,8 @@ public ActionResult<StatusResponseModel> GetStatus()
         return new StatusResponseModel
         {
             Commit = _opdexConfiguration.CommitHash,
-            Identifier = _opdexConfiguration.InstanceId
+            Identifier = _opdexConfiguration.InstanceId,
+            UnderMaintenance = _maintenanceConfig.Value.Locked
```
mrtpain, line 34: Wondering if we should move, or also add this to the Index status endpoint.

Also considering preventing Index if either `Available` is false or if `Maintenance` is true. What do you think?

mrtpain (approved):
One comment to consider, otherwise lgtm

PR #231, opened 2022-03-10
[PAPI-399] Reduce AKV reload interval

drmathias:
The long reload interval makes it difficult to diagnose key vault caching issues. Reducing it won't be a problem because it doesn't consume a lot of resources and won't even get close to any Azure limits. Every 16 seconds matches with block time, so config changes _should_ get propagated for the following block.


mrtpain:
I don't think this will change anything, when I tested I waited 20 minutes for maintenance lock to refresh and it never did.

mrtpain (approved):

PR #241, opened 2022-04-14
[PAPI-398] Update CoinGecko client

drmathias:
Resolves bug where price data was failing to deserialise if CoinGecko was used as pricing feed


drmathias:
Doesn't seem to actually resolve the issue :(

PR #242, opened 2022-04-14
[PAPI-421] Local call retries

drmathias:
Retries local calls when there's a "No code at this address." response from the node. Uses an backoff of 500ms -> 1s -> 1.5s


src/Opdex.Platform.Infrastructure/Clients/CirrusFullNodeApi/Modules/SmartContractsModule.cs
```diff
@@ -77,7 +78,21 @@ public async Task<LocalCallResponseDto> LocalCallAsync(LocalCallRequestDto reque
...
+                {
+                    await Task.Delay(TimeSpan.FromMilliseconds(500 + (500 * retryCount)), cancellationToken);
+                    continue;
+                }
+
+                response = callResult;
+                break;
+            }
+
+            return response;
```
mrtpain, line 95: Rather than setting `callResult` we might want to just set `response = await PostAsync...` If we fail 3 times in a row, we're going to return null which is going to cause null reference errors.

So 3 failures will still return the failed local call response with the error message

drmathias: Sorted

mrtpain (approved):

PR #249, opened 2022-05-19
[PAPI-426] Proposal pledge paging use alias with column name

drmathias:
Resolves bug where paging of proposal pledges returns a 500 Internal Server Error


mrtpain (approved):
😅

